PR #19, opened 2023-02-09
MERL-544: add block type to wpgraphql

matthewguywright:


includes/Blocks/Block.php
```diff
@@ -61,10 +61,23 @@ public function __construct( WP_Block_Type $block, Registry $block_registry ) {
...
 		$this->register_block_type();
 	}
 
+	/**
+	 * Update scaler types
+	 * 
+	 * @return void
+	 */
+	public function update_scalars() {
+		Registry::register_graphql_scalar( 'BlockAttributesObject', [
```
theodesp, line 74: There is no `Registry::register_graphql_scalar` call and the test cases are failing. According to the [documentation](https://www.wpgraphql.com/functions/register_graphql_scalar) it should be `register_graphql_scalar`

includes/Blocks/Block.php
```diff
@@ -61,10 +61,23 @@ public function __construct( WP_Block_Type $block, Registry $block_registry ) {
 		$this->block_registry   = $block_registry;
 		$this->block_attributes = $this->block->attributes;
 		$this->type_name        = $this->format_type_name( $block->name );
-
+		$this->update_scalars();
```
theodesp, line 64: Also I would not put this method inside the block type as this gets called on every Block we register. Ideally you want to create a new class that inherits the  `OnInit` interface and call this inside the Registry `OnInit` method:

https://github.com/wpengine/wp-graphql-content-blocks/blob/main/includes/Registry/Registry.php#L52-L56

theodesp (approved):
LGTM!

PR #20, opened 2023-02-13
Add support for `UnknownBlock` type

blakewilson:
Adds an `UnknownBlock` type to the resolver for blocks that are not supported in WordPress.


includes/Blocks/UnknownBlock.php
```diff
@@ -0,0 +1,24 @@
+<?php
+namespace WPGraphQL\ContentBlocks\Blocks;
+
+class UnknownBlock extends Block {
+	protected ?array $additional_block_attributes = array(
+		'style'        => array(
```
theodesp, line 6: You may want to remove those properties as they may not exist in the real block if you just want to inherit from `Block`

theodesp: You also do not need this `class UnknownBlock` as you are only interested in the 
`register_graphql_object_type` that you have below. So you can remove this file.

blakewilson: Great, file removed!

theodesp (approved):

PR #21, opened 2023-02-19
dev: add package type to composer.json

justlevine:
This PR sets the Composer package type to "wordpress-plugin", so dependents can autoinstall the plugin in `wp-content/plugins` directory using the `extra.installer-paths` property.


montchr:
@justlevine FYI you added the change to `package.json` not `composer.json` 😉

justlevine:
@montchr good looking out! I should rly know better than to fork+commit from my phone in line at the grocers 😳

Forced push the change

theodesp (approved):

PR #23, opened 2023-02-21
Chore: Remove Extra Block `style` fields

theodesp:
This PR removed the erroneous `style` fields added on some blocks as they create issues with WPGraphQL type checking.


TeresaGobble (approved):
Good catch, thank you Theo!

PR #34, opened 2023-03-10
fix: illegal index error fixed in Block.php

kidunot89:
# Summary

image

I noticed this error in my Docker console as I would run queries with `editorBlocks` in them, so I just moved the assignment of `$default` above the `if` statement in the `Block::resolve_block_attributes()` function, replaced the following usages of `$attribute_config['default']` with `$default`.


theodesp:
@kidunot89 Awesome. I will take a look.

theodesp (approved):

PR #35, opened 2023-03-14
Bundle `vendor` in the final plugin .zip (Remove Composer step)

blakewilson:
This PR removes the `composer install` step by zipping the final plugin dist and uploading it to the GitHub release upon a successful Changesets release.

Users will be able to visit the [releases](https://github.com/wpengine/wp-graphql-content-blocks/releases) tab and get the source code as they could before, which is an exact representation of the GitHub repo, as well as download the `wp-graphql-content-blocks.zip` plugin file, which includes the prod `vendor` directory, and all other [required dist files as specified in the package.json.](https://github.com/wpengine/wp-graphql-content-blocks/blob/remove-compser-step/package.json#L22-L29)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,6 +25,8 @@ jobs:
         id: changesets
         uses: changesets/action@v1
         with:
+          # This expects you to have a script called release which does a build for your packages and calls changeset publish
+          publish: npm run release
           version: npm run version
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,6 +40,6 @@ jobs:
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: "@wpengine/wp-graphql-content-blocks.zip"
-          asset_name: wp-graphql-content-blocks.zip
+          asset_name: ${{ format('wp-graphql-content-blocks.{0}.zip', fromJSON(steps.changesets.outputs.publishedPackages)[0].version) }}
           tag: ${{ format('v{0}', fromJSON(steps.changesets.outputs.publishedPackages)[0].version) }}
           overwrite: true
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -25,6 +25,21 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload release artifact
+        if: steps.changesets.outputs.published == 'true'
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: "@wpengine/wp-graphql-content-blocks.zip"
+          asset_name: wp-graphql-content-blocks.zip
```
josephfusco, line 43: It could be helpful to add the version directory to the filename to help identify the download without needing to unzip.

josephfusco (approved):
Just one non-blocking piece of feedback

README.md
```diff
@@ -97,49 +93,53 @@ For example, given the following HTML Content:
...
+      "className": null
+    }
+  }
 ]
 ```
 
 The `CoreColumns` contains one or more `CoreColumn` block, and each `CoreColumn` contains a `CoreParagraph`.
 
-Given the flattened list of blocks though, how can you put it back? Well that's where you use the `` and `parentId` fields to assign temporary unique ids for each block.
+Given the flattened list of blocks though, how can you put it back? Well that's where you use the ``and`parentId` fields to assign temporary unique ids for each block.
```
theodesp, line 136: ``and`parentId` should be and `parentClientId`. I had the impression that we have updated the README for the new changes but apparently not.

.github/workflows/release.yml
blakewilson: Good call! I've updated the asset name to be formated like `wp-graphql-content-blocks.0.1.1.zip` to replicate how Wordpress does it on the plugins repo:

https://downloads.wordpress.org/plugin/faustwp.0.8.1.zip

README.md
blakewilson: Good catch! Fixed.

josephfusco (approved):

PR #42, opened 2023-03-23
Change artifact name and readme links

blakewilson:
As @TeresaGobble begins working on a way to get automatic updates from GitHub, we need a consistency named .zip file for the updater to grab. Changing the artifact name to `wp-graphql-content-blocks.zip`.


TeresaGobble (approved):
Thank you for this, Blake!

Issue #15, opened 2023-01-30
General Feedback on Schema Design


jasonbahl:
This is some general feedback after initial tinkering with the plugin. 

First off, good stuff! This is working similar to what I had in mind when I began work on [WPGraphQL Block Editor](https://github.com/wp-graphql/wp-graphql-block-editor). 🙌🏻 

I have some feedback I think we should consider, ideally before we get too many users as they would be breaking changes. 

### Rename ContentBlock Interface

Not necessarily a hill I will die on, but I think `Content` usually has connotations with things like blog posts, but nav menus (in my mind) are not the same as "content". 

I had selected "[EditorBlock](https://github.com/wp-graphql/wp-graphql-block-editor/blob/master/src/Type/InterfaceType/EditorBlockInterface.php#L103)" as the name of the interface because each block is a block of the block editor. It _might_ be a ContentBlock, or it _might_ be a menu block or a layout block, etc. 

I think we can do users (and ourselves) a service by being more clear with the naming. 

Maybe something like: 

- EditorBlock: All blocks that can be used within the Block Editor
- ContentBlock (implements EditorBlock): Blocks used to manage content
- LayoutBlock (implements EditorBlock): Blocks used to manage layout
- NavMenuBlock(?) (Implements EditorBlock): Blocks used by the block editor when managing nav menus

By having a generic (EditorBlock) Interface, and then more specific Interfaces (Content, Layout, etc) we can better use the Schema as documentation, and provide better experiences for our users. 

We can even get more specific like: 

- PostContentBlock (Implements ContentBlock, EditorBlock): Blocks used to manage content on the "Post" type
- PageContentBlock (Implements ContentBlock, EditorBlock): Blocks used to manage content on the "Page" type

etc. 

This way, we can allow the Schema to better differentiate for consumers what blocks they can expect. 

For example, if I register a custom block _only_ on the `Car` post type, It shouldn't show up as a possible block on the `Page` and `Post` post types, because we _know_ the block will never be available on those types. 

So, if our schema for posts was: 

```graphql

# A block used by the block editor
Interface EditorBlock {
  name: Name of the block
  clientId: Non-persisted ID used to map blocks by the client application
  clientParentId: Non-persisted ID of the parent block used to map hierarchical relationships in the client application
}

# Editor Block used to manage content
Interface ContentBlock implements EditorBlock {
  name: Name of the block
[...]

PR #47, opened 2023-04-03
Implements support for block type filter

theodesp:
This PR creates additional schema changes based on the result of the `get_allowed_block_types` filter.

More specifically:

* If the result of the `get_allowed_block_types` filter returns a list of block names, then we register specialized interfaces for each block type.
For example if `get_allowed_block_types` returns  `['core/paragraph', 'core/image']` for the `post` type then it registers the following schemas:

```
interface PostBlock implements EditorBlock {};

Interface NodeWithPostBlocks implements NodeWithEditorBlocks {
  editorBlocks: [PostBlock];
}
```
* Refactors the code to make the logic in the Registry class more slimmed down.
* Filters out the list of resolved blocks from the response based on the `get_allowed_block_types`. For example if the post content will contain existing blocks not from the allow list they will be filtered out from the GraphQL query.

## How to Test

* Check that by default if you do not provide a filter `get_allowed_block_types` then no schema changes are observed. So there should be no `NodeWithPostBlocks` or `PostBlock` in the schema for example.

* Provide a custom `get_allowed_block_types` filter. Here is an example:
```
add_filter( 'allowed_block_types_all', 'set_allowed_block_types', 10, 2 );

/**
 * Set allowed gutenburg block types
 * https://rudrastyh.com/gutenberg/remove-default-blocks.html
 */
function set_allowed_block_types( $allowed_blocks, $editor_context ) {

	if ( isset( $editor_context->post ) && $editor_context->post instanceof WP_Post && 'post' === $editor_context->post->post_type ) {
		return  [
		   'core/image',
		   'core/paragraph'
		];
	 }
	}
	return true;
}
```

With that filter in place you should be able to query the following fields:

* editorBlocks field for posts should be of type `[PostBlock]`
Screenshot 2023-04-03 at 13 57 35

* editorBlocks field for page should be of type `[EditorBlock]`
Screenshot 2023-04-03 at 13 59 12

* `PostBlock` implementation should only include the `CoreImage` and `CoreParagraph` blocks.
Screenshot 2023-04-03 at 14 00 48

* `EditorBlock` implementation should only include all the available blocks.

* There is a new interface `NodeWithPostBlocks` containing a list of `PostBlock` types:
Screenshot 2023-04-03 at 14 03 13

* The list of resolved block data only includes blocks from the allow list. To test that you need to remove the `allowed_block_types_all` filter, create a post with a block not in the previous list  and then enable the filter again. You should be able to see only the filtered blocks:

**Before:**
Screenshot 2023-04-03 at 14 08 22

**After:**
Screenshot 2023-04-03 at 14 09 08

**Note**. The reason we filter out the blocks from the response is that the resolveType will fail at runtime if we provide a block not implementing the `PostBlock` interface.
After conversation with **jasonbahl** he suggested that there shouldn't be any `UnknownBlock` in case the content resolves a block not in the allow list but instead removed completely.  The caveat here is that it could create some layout blocks not to appear, for example: User has `CoreColumns` in posts. User removes `CoreColumns` from the allowed block list. Then the inner blocks will be orphaned since there is no parent block to host the child blocks.

## Screen Shots


includes/Registry/Registry.php
```diff
@@ -51,56 +54,69 @@ public function __construct( TypeRegistry $type_registry, $block_type_registry )
...
 
 	/**
-	 * This adds the WP Block Registry to AppContext
+	 * Register Interface types to the GraphQL Schema
 	 *
 	 * @return void
 	 */
-	public function pass_blocks_to_context() {
-		add_filter( 'graphql_app_context_config', array( $this, 'load_registered_editor_blocks' ) );
+	protected function register_interface_types() {
```
theodesp, line 68: All logic is contained in this method.

justlevine:
>With that filter in place you should be able to query the following fields:
>- editorBlocks field for posts should be of type `[PostBlock]`
>- editorBlocks field for page should be of type `[EditorBlock]`

Shouldnt it be `{$PostType}Block implements EditorBlock` for _all_ types, including pages?

Pages are no different than any other Post Type, so even putting aside most the other feedback in #15 , this approach means you cant specifically target page blocks like you can Posts or any other CPT.

includes/Data/ContentBlocksResolver.php
```diff
@@ -66,9 +66,8 @@ function ( $parsed_block ) {
 			},
 			$parsed_blocks
 		);
-
 		// Flatten block list here if requested or if 'flat' value is not selected (default)
-		if ( !isset( $args['flat'] ) || 'true' == $args['flat'] ) {
+		if ( !isset( $args['flat'] ) || 'false' == $args['flat'] ) {
```
justlevine, line 70: Is this correct? I know the default behavior got changed in a prev PR, but I think `$args['flat']` still refers to a user _wanting_ to flatten the array.

theodesp:
> 

Yeah I think that is the case:

The `PostBlock` implements `EditorBlock` but in the Schema you will see the type as `[PostBlock]`.

theodesp:
Shouldn't the hierarchy be:

```
# A block used by the block editor
Interface EditorBlock {
  name: Name of the block
  clientId: Non-persisted ID used to map blocks by the client application
  clientParentId: Non-persisted ID of the parent block used to map hierarchical relationships in the client application
}
# A block used by the block editor for Post Type
Interface PostBlock implements EditorBlock {}

# A block used by the block editor for Page Type
Interface PageBlock implements EditorBlock {}

# List of blocks used by the block editor
Interface NodeWithEditorBlocks {
   editorBlocks: [EditorBlock]
}  
# List of blocks used by the block editor associated with the Post post type
Interface NodeWithPostBlocks implements NodeWithEditorBlocks {
   editorBlocks: [PostBlock]
} 

# List of blocks used by the block editor associated with the Page post type
Interface NodeWithPageBlocks implements NodeWithEditorBlocks {
   editorBlocks: [PageBlock]
}
```

justlevine:
> Shouldn't the hierarchy be like:


Thats **exactly** what I'm hoping for. Is that how the PR is expected to work?

Perhaps I misunderstood what you meant in the PR description when you wrote "editorBlocks field for page should be of type `[EditorBlock]`" in the initial PR description? I understood that to mean that we would would _not_ be getting `NodeWithPageBlocks`/ `[PageBlock]` interfaces.

theodesp:
> editorBlocks

Yes for the example That I use above with the filter;

```
function set_allowed_block_types( $allowed_blocks, $editor_context ) {

	if ( isset( $editor_context->post ) && $editor_context->post instanceof WP_Post && 'post' === $editor_context->post->post_type ) {
		return  [
		   'core/image',
		   'core/paragraph'
		];
	 }
	}
	return true;
}
```

Since we are returning `true` for Page blocks for example then it will not be getting interfaces for `NodeWithPageBlocks/ [PageBlock] interfaces` since they are implied to use the default `NodeWithEditorBlocks` and `[EditorBlock]` interfaces.

justlevine:
Ah okay, thanks for clarifying! I was missing that the interface _only gets created/applied conditionally_.

I think I would still recommend we **always** create and apply `NodeWith{PostType}Blocks`, even if there is currently no default difference between `{PostType}Block` and `EditorBlock`. This way devs and component creators can create reusable code/graphql fragments and target/consume those interfaces without worrying about breaking schema changes based on what blocks a site might have enabled.

theodesp:
> Ah okay, thanks for clarifying! I was missing that the interface _only gets created/applied conditionally_.
> 
> I think I would still recommend we **always** create and apply `NodeWith{PostType}Blocks`, even if there is currently no default difference between `{PostType}Block` and `EditorBlock`. This way devs and component creators can create reusable code/graphql fragments and target/consume those interfaces without worrying about breaking schema changes based on what blocks a site might have enabled.

I had the same question so I asked @jasonbahl  and he thinks it's better to leave it as it is:
> The way WPGraphQL Core works today is that it uses the registries (Post Type, Taxonomy, etc) that exist in WordPress core and maps those to the GraphQL Schema. So as you add or remove things from the registry, the Schema will change.

> This should be the same for Blocks.

> If I register new post types with support for blocks, I should get a new interface(s) that represent this new functionality.
If I remove post types or blocks, those types should be removed from the Schema as the system will no longer fulfil those queries, etc.

justlevine:
@theodesp @jasonbahl the difference _here specifically _ is that block support isn't what is changing changing, but rather the existence of specific blocks on the post type. 

The parallel would be if WPGraphQL Core removed the `PostType` or `Taxonomy` types/interfaces if the site had _no posts or terms registered_** instead of if the post type or taxonomy doesn't exist at all on the instance.

jasonbahl:
Ya, so I think the Schema should represent what's _actually_ possible. 

If it's possible for blocks to be added to a post type, those blocks should be represented in the Schema, but only the ones that are possible to be added to a particular post type. 

Let's say we had some specific blocks that were registered _only_ for the `Product` post type. Those blocks should implement the `ProductBlock` interface and be possible types of the `editorBlocks` field on the `Product` type. 

But they should not implement the `PostBlock` interface because they are not available on Posts for insertion in the Admin or querying via GraphQL.

If a post type has restricted Gutenberg altogether, only allowing the classic editor, for example, that post type should NOT implement the `NodeWith{$TypeWithoutBlocks}Blocks` interface, because that post type doesn't have the possibility of having blocks added to posts of that type or querying blocks on posts of that type.

If a post type technically has gutenberg enabled, but has (for some reason) filtered out (deny-listed) every single block, then technically the post type would not have block support as the Admin UI wouldn't allow users to insert blocks if zero blocks were allowed for insertion, and in that case I would argue the Schema should match that and not show the `NodeWith{$TypeWithZeroPossibleBlocks}Blocks` Interface.

I'm not actually sure what happens in the Post Edit screen if you deny-list ALL block types. But I assume it would be weird. Possibly it falls back to classic editor too? Not sure how one would use Gutenberg if there were zero blocks allowed 🤔

justlevine:
@jasonbahl Agree with everything you wrote.

> If a post type technically has gutenberg enabled, but has (for some reason) filtered out (deny-listed) every single block, then technically the post type would not have block support as the Admin UI wouldn't allow users to insert blocks if zero blocks were allowed for insertion, and in that case I would argue the Schema should match that and not show the NodeWith{$TypeWithZeroPossibleBlocks}Blocks Interface.

My understanding here is that the **opposite** is happening. `EditorBlock` is the parent interface, and is applied to all block-supporting CPT types by default.  `{PostType}Block` is the implementing interface for `EditorBlock` items on a particular PostType.

So far so good.

The problem is that instead of creating that interface always and  _targeting_ that interface to filter the allowed blocks (in this PR, but not the sole semantic use for the Interface), it instead _only_ creates the interface (and changes the return type!) if the list of allowed block types differs from the default list. 

Which means that instead of just being able to write:

```graphql
`Fragment myFrag on ContentNode {
  ... on NodeWithEditorBlocks {
    editorBlocks {
      ... someCommonFrag
      ... on PostBlock {
        ... someProductContextFrag
      }
    }
  ... on NodeWithPageBlock {
    ...somePageContextFrag
    editorBlocks {...}
  }
}
```

that would cause the schema to error out, when semantically it should work no differently than doing `contentNodes { ... on Page {...} }` when no pages have yet to be created.

theodesp:
Hey @justlevine  I can see that happening. There are pros and cons of each approach.

It may look inconvenient but it maybe useful for the schema to error if you are using an interface that doesn't make sense. For example if you change something in your `set_allowed_block_types` filter that makes significant changes to the API.

On the other hand, you want to have a safe bet that even if you are querying using `NodeWithPageBlocks` you will not get an error if you decide one day to remove the `set_allowed_block_types` filter.

However keeping it would mean that the schema now is  noticeable bigger, bulkier and a bit more confusing to navigate since it will contain duplicate models.

There is a small escape hatch here in the code. Notice that I used the following check:
```
if ( is_array( $supported_blocks_for_post_type ) ) {
```
So technically you can return an empty array for each post type and still get a `NodeWith[PostType]Blocks` interface even though it has 0 blocks there

So based on those thoughts, we can either:
* Make it more strict and accurate: Only registering  `NodeWith[PostType]Blocks` when needed.
* Make it more permissive: Allowing all `NodeWith[PostType]Blocks` even if they don't have blocks

justlevine:
@theodesp personally in this particular case I don't believe the interface should error out because it's still semantically valid. 
Unless a post type doesn't support blocks at all, then has the shape of `NodeWith{PostType}Blocks`, which means that interface should _always_ be available for targeting. Anything less imo is going to be a nightmare for extensibility and reusability.
Implementing interfaces don't appear in the GraphiQL Explorer unless they're written in the query (thanks to lazy loading), so the n+2 interfaces added to the schema I don't think are at the level of mitigating the schema benefits of these types.

That said, now that we're at least on the same page about what the issue here is, I'ma shut my mouth and defer to you.

tests/unit/RegistryTestCase.php
```diff
@@ -15,66 +15,162 @@ public function setUp(): void {
...
+		add_filter(
+			'allowed_block_types_all',
+			function ( $allowed_blocks, $editor_context ) {
+				if ( isset( $editor_context->post ) && $editor_context->post instanceof \WP_Post && 'post' === $editor_context->post->post_type ) {
+					return array(
+						'core/image',
+						'core/paragraph',
+					);
+				}
+				return true;
```
blakewilson, line 79: So if the filter API can either return a `string[]` of blocks, or `true` for all?

.changeset/lovely-doors-hammer.md
blakewilson, line 1: Not sure what happened here, but theses changes should already be in `main`. Maybe a rebase is needed?

includes/Blocks/CoreImage.php
```diff
@@ -5,7 +5,7 @@
 class CoreImage extends Block {
 
 	protected ?array $additional_block_attributes = array(
-		'className' => array(
+		'cssClassName' => array(
```
blakewilson, line 8: Was this a bug? Do we need a changeset here?

blakewilson (approved):
Nice work @theodesp! Followed the testing instructions and it works perfectly.

TeresaGobble (approved):
Before and after- everything looks good! Thank you Theo! Folks will get so much use out of this!
image
image

image

jasonbahl:
> The problem is that instead of creating that interface always and targeting that interface to filter the allowed blocks (in this PR, but not the sole semantic use for the Interface), it instead only creates the interface (and changes the return type!) if the list of allowed block types differs from the default list.

> Which means that instead of just being able to write:

```graphql
fragment myFrag on ContentNode {
  ... on NodeWithEditorBlocks {
    editorBlocks {
      ... someCommonFrag
      ... on PostBlock {
        ... someProductContextFrag
      }
    }
  ... on NodeWithPageBlocks {
    ...somePageContextFrag
    editorBlocks {...}
  }
}
```

> that would cause the schema to error out. Semantically it should work no differently than doing contentNodes { ... on Page {...} } when no pages have yet to be created.

-----

@justlevine I _think_ the better analogy here would be: 

```graphql
fragment myFrag on ContentNode {
  ... on Page {
    ...
  }
  ... on Post {
   ...
  }
  ...on SomeNewCustomPostType {
     ...
  }
}
```

The `...on SomeNewCustomPostType` will error if that post type hasn't been registered to the schema. 

Much like `...on NodeWithSomeNewCustomPostTypeBlocks` would error if that post type wasn't registered. 

Same would be for `...on PageBlocks` if the `Page` type has had Gutenberg turned off / all block types removed from that post type. 

If it's truly not possible for blocks to be queried for a post type, the Schema shouldn't attempt to express that it's possible. 

Sure this might lead to errors, but it can also lead to tooling better helping you understand if/when something has gone wrong. 

For example, via Schema difs etc we could determine at what point in time a specific Post Type stopped supporting gutenberg because the Schema no longer had `NodeWith{$TypeName}Blocks` interface. 

Perhaps you didn't mean to disable Gutenberg for the post type and tooling would be able to tell you about the breaking change to the schema. 

I `NodeWith{$TypeName}Blocks` existed even when Gutenberg was de-activated for that post type, the schema would be misleading, indicating that that Type does have blocks (it doesn't) and might actually make it more difficult to debug why no blocks are being returned (hint the resolver shouldn't return _any_ blocks for a post type that doesn't support any blocks)

justlevine:
@jasonbahl to clarify, by `always` I mean "any time `NodeWithEditorBlocks` is applied to a type" vs the current approach where it's applied "whenever the allowlist of blocks differs from primary implementation" and decidedly not when a type has _zero_ allowed blocks.  I'd argue in that case a type should have _neither_ `NodeWithEditorBlocks` nor `NodeWith{PostType}Blocks`, but since it doesnt break the predictability or semantic meaning of the schema, I'm less bothered by that then by  using an interface as a faux union/filtered connection.

Messaged you on slack to get a bit of additional clarification and hopefully prevent me from taking us all in circles again.

theodesp:
Let's keep it stricter for now since we have done most of the work and it would be more clear if the schema changes, then the clients would be aware of the change.

I will also add a ticket for CI/CD pipeline to detect breaking schema changes based on https://www.graphql-cli.com/diff/

jasonbahl:
@theodesp we should address this now as it will be a breaking change to address later.

jasonbahl:
I can probably write some tests if it would be helpful.

jasonbahl:
Ok, so I pulled the code down and "fixed" it so that the `NodeWith${TypeName}Blocks` is applied to any post type that has block support. 

But now the Schema takes 45s to generate. 

I _think_ there might be some nested loops happening that are causing things to execute a lot more often than maybe necessary

jasonbahl:
Let's say I wanted to remove the "Quote" block from all post types. 

I could do that with the following snippet: 

```php
add_filter( 'allowed_block_types_all', 'test_remove_quote_block', 10, 2 );

function test_remove_quote_block( $allowed_blocks, WP_Block_Editor_Context $editor_context ) {

	if ( ! isset( $editor_context->post->post_type ) ) {
		return $allowed_blocks;
	}

	if ( $editor_context->post->post_type !== 'page' ) {
		$all_blocks = \WP_Block_Type_Registry::get_instance()->get_all_registered();
		unset( $all_blocks['core/quote'] );
		return array_keys( $all_blocks );
	}

	return $allowed_blocks;
}
```

This will remove the `core/quote` block from all post types _other_ than page. 

With this snippet in place, the Schema is now taking me ~45-50s to load:

CleanShot 2023-04-07 at 11 53 51

I do get the `PostBlock`, etc Interfaces, but I think there might be some sort of recursion issue happening. 🤔

includes/utilities/WPHelpers.php
```diff
@@ -0,0 +1,74 @@
...
+	 * @param string $post_type The Post Type to use.
+	 * @param number $id The Post Id to use.
+	 *
+	 * @return WP_Block_Editor_Context The Block Editor Context
+	 */
+	public static function get_block_editor_context( $post_type, $id = -99 ) {
+		$post_id              = $id;
+		$post                 = new stdClass();
+		$post->ID             = $post_id;
+		$post->post_author    = 1;
```
jasonbahl, line 59: @theodesp when using a plugin like Block Visibility, I believe this will cause the Schema to be conditional based on blocks allowed by whoever the User with ID 1 is, and I think that will cause problems. 

We should probably set the `post_author` to an invalid ID as well, similar to the post id.

theodesp: Sure I will take a look

jasonbahl:
> With this snippet in place, the Schema is now taking me ~45-50s to load

If I comment out lines 76-108 of Registry.php, the Schema takes 9s to load instead of ~45s.

I think the slowdown might be the [array_map](https://github.com/wpengine/wp-graphql-content-blocks/pull/47/files#diff-ba0150e300f9b1241e52cdf70c23f1da495d595cf592712b91818a091093b806R95)  inside a  [foreach](https://github.com/wpengine/wp-graphql-content-blocks/pull/47/files#diff-ba0150e300f9b1241e52cdf70c23f1da495d595cf592712b91818a091093b806R79) loop. 

🤔 

I'll tinker with this and see if maybe there's a more efficient way of accomplishing this.

PR #49, opened 2023-04-07
Handle more block attribute sources

blakewilson:
## Description

* Added support for `html` source block attributes with the `multiline` property.
* Added support for `text` source block attributes
* Added support for `integer` type block attributes


matthewguywright (approved):

josephfusco (approved):
LGTM!

PR #52, opened 2023-04-07
fix: slow schema / slow queries / unexpected Schema output

### LINKED ISSUE #51: regression: slow schema load time, query execution time
After merging PR https://github.com/wpengine/wp-graphql-content-blocks/pull/47, the schema is slow to load and queries are slow to execute. Additionally, all post types `editorBlocks` have the return type `[ EditorBlock ]` instead of `[ ${PostType}Block ]` as was the intent of #47. 

## Steps to Reproduce:

1. Setup a WordPress environment running WPGraphQL v.1.14.0 and the latest "main" branch of WPGraphQL Content Blocks with the above PR merged
2. Load the Schema in GraphiQL and note the time it took. 
3. Add the following snippet to remove support for the "core/quote" block from Gutenberg

```php
add_filter( 'allowed_block_types_all', 'test_remove_quote_block', 10, 2 );

function test_remove_quote_block( $allowed_blocks, WP_Block_Editor_Context $editor_context ) {

	if ( ! isset( $editor_context->post->post_type ) ) {
		return $allowed_blocks;
	}

	if ( $editor_context->post->post_type !== 'page' ) {
		$all_blocks = \WP_Block_Type_Registry::get_instance()->get_all_registered();
		unset( $all_blocks['core/quote'] );
		return array_keys( $all_blocks );
	}

	return $allowed_blocks;
}
```

4. Load the schema again in GraphiQL and note that it's taking significantly longer. (For me it was 9s before adding the snippet above  and 45s after). 

Also, without this snippet above, I don't get the `editorBlocks: [ PageBlock ]`, `editorBlocks: [ PostBlock ]` in the Schema. . . all post types have the `editorBlocks: [ EditorBlock ]` as the field return type. This isn't expected behavior, as discussed in the comments of #47. I would expect each Post Type (and other block context locations in the future) to have the blocks represented in the schema with a return type of `[ $typeName . 'Block' ]` so that we can more accurately understand what blocks are available on each type (the intention behind much of the comments in #15).

5. Even queries such as `{posts{nodes{id}}}` are slower after adding the snippet above

----

## Proposed Solution

I believe the culprit comes down to this section of code:  https://github.com/wpengine/wp-graphql-content-blocks/pull/47/files#diff-ba0150e300f9b1241e52cdf70c23f1da495d595cf592712b91818a091093b806R76-R108

Most specifically line #103. 

Right now there's an array_map within a foreach loop causing an N+1 problem. For each post type registered to show in GraphQL there will be an array_map passed over all blocks associated with that post type. I'm not sure that's a significant issue, but one that might be better handled once instead of N+1 times. 

Line #103 attempts to register a lot of interfaces on a lot of Types for EVERY single GraphQL request. I don't think this is necessary. 

Instead of adding Interfaces to the Block Types here, I think we can more efficiently add the interfaces at the time of registering the Block Types. 

What I propose is the Registry collect (one time and memoize it) a list of Blocks supported on each post type. 

Then, when the Block Type is being registered and interfaces are being applied to each Block Type, it can reference that list and apply the necessary interfaces at that time. 

> Even queries such as `{posts{nodes{id}}}` are slower after adding the snippet above

This should help only generate the Interfaces for GraphQL requests that need them to be fulfilled, vs generating them on every request. 

Queries like that shouldn't be needing to generate Interfaces for PageBlocks, CarBlocks, HouseBlocks, etc. . .becuase they weren't queried for. But right now each GraphQL request is paying the price of the overhead of registering the interfaces on every request.

jasonbahl:
This PR aims to close #51. 

This PR does the following:

- Block.php
- - introduce "get_block_interfaces"
- Registry.php
- - introduce get_block_context_interfaces method, which will populate the $block_interfaces class variable with an array of interfaces per block type and will return a specific array of interfaces given a block name
- - removed the call to `register_graphql_interfaces_to_types()` that was registering interfaces to all block types iteratively.
- - renamed `NodeWith . $post_type . 'Blocks'` to `NodeWith. $post_type . 'EditorBlocks' to keep naming more consistent
- PostTypeBlockInterface.php
- - renamed `NodeWith . $post_type . 'Blocks'` to `NodeWith. $post_type . 'EditorBlocks' to keep naming more consistent
- - renamed `ucfirst( $post_type ) . 'Block'` to `ucfirst( $post_type ) . 'EditorBlock'` to keep naming more consistent
- ran composer phpcs:fix

----

## Before (with main #47 merged)

Loading the Schema in GraphiQL takes 40 seconds:

CleanShot 2023-04-07 at 15 07 37

Query for posts (not even asking for blocks) takes +10 seconds:

CleanShot 2023-04-07 at 15 12 52

The "Page.editorBlocks" field returns a listOf EditorBlock, but I expect this to be a listOf `PageEditorBlock`

CleanShot 2023-04-07 at 15 22 03

## After (changes from this PR applied)

Loading the Schema in GraphiQL takes 9 seconds:

CleanShot 2023-04-07 at 15 14 40

Executing a query for posts not even asking for blocks) takes 1.7 seconds:

CleanShot 2023-04-07 at 15 15 47

The "Page.editorBlocks" field returns a listOf PageEditorBlock, as expected (instead of listOf EditorBlock)

CleanShot 2023-04-07 at 15 18 08

### DIFF
--- a/includes/Registry/Registry.php
+++ b/includes/Registry/Registry.php
@@ -36,6 +36,8 @@ final class Registry implements OnInit {
 	 */
 	public $registered_blocks;
 
+	public $block_interfaces = array();
+
 	/**
 	 * Registry constructor.
 	 *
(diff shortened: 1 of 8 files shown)


includes/Registry/Registry.php
```diff
@@ -59,12 +61,89 @@ public function OnInit() {
...
+			// Push the Contextual Interface for each supported Block to the block_interfaces array
+			array_map(
+				function( $supported_block ) use ( $type_name ) {
+					$this->block_interfaces[ $supported_block ][] = Utils::format_type_name( $type_name . 'EditorBlock' );
+				},
+				$supported_blocks_for_post_type_context
+			);
+		}//end foreach
+
+		return $this->block_interfaces[ $this->block->name ] ?? array();
```
theodesp, line 121: This line breaks tests since `$this->block->name` is not defined. Is this supposed to be `$block_name` instead?

theodesp (dismissed):
LGTM! Awesome!

theodesp (approved):

blakewilson:
Nice work @jasonbahl! Thanks for the PR. Let's make sure we also get a changeset for this.

jasonbahl:
oh, the changeset bot already tried to give me a link 🤣

jasonbahl:
@blakewilson ya, following the changeset instructions (I tried with npx and yarn) I get the following: 

```bash
➜  wp-graphql-content-blocks git:(fix/#51-slow-schema-and-missing-interface-return-types) ✗ npx changeset
npm ERR! could not determine executable to run

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/jason.bahl/.npm/_logs/2023-04-10T21_01_43_075Z-debug.log
➜  wp-graphql-content-blocks git:(fix/#51-slow-schema-and-missing-interface-return-types) ✗ yarn changeset
yarn run v1.22.18
warning ../../package.json: No license field
$ changeset
/bin/sh: changeset: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

jasonbahl:
There's a failing test at the moment (update coming) because the test is incorrect. 

When you register a post_type as follows (in the test): 

```php
register_post_type(
		'recipe',
		array(
			'show_in_graphql'     => true,
			'graphql_single_name' => 'recipe',
			'graphql_plural_name' => 'recipes',
			'supports'            => array( 'title', 'editor', 'author', 'thumbnail' ),
			'public'              => true,
		)
	);
```

the post type will NOT be block editor enabled. See: 

CleanShot 2023-04-10 at 15 03 47

The test is asserting that this post type should be a supported post type for WPGraphQL Block Editor, but it should not be, because it's not block-editor enabled.

jasonbahl:
Pushed up another commit fixing the test: `test_get_supported_post_types`

theodesp:
Interesting I had the impression that because it has the 'editor' support it should be block editor enabled.

jasonbahl:
> Interesting I had the impression that because it has the 'editor' support it should be block editor enabled.

@theodesp you would think it would be something like: 

```php
'supports' => [ 'block_editor' ]
```

but for whatever reason it is much less straight-forward than that. 🤷🏻‍♂️

PR #59, opened 2023-04-27
[Merl-862] Add support for Classic Blocks

theodesp:
# Description

This PR adds support for querying Classic Blocks. Since those blocks are not stored in the database with a proper name we assign them a name now and resolve them as `CoreFreeForm` returning the contents of the innerHTML as long as the contents are not empty.

# Screenshots
Screenshot 2023-04-27 at 13 07 58

# Testing
1. Create a Classic block in a post
2. Query for the `content` attribute in editor blocks.
4. Block should resolve as `CoreFreeForm` as plain HTML.

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -73,7 +73,7 @@ class Block {
 	 * @return void
 	 */
 	public function register_fields() {     }
-	
+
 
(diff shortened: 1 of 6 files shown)


josephfusco (approved):
Looks good!

Issue #62, opened 2023-05-03
Block content and inner blocks "anchor" field being applied to parent block resulting in duplicates


kidunot89:
The current selector for the "anchor" field is invalid. It keeps looking for an `id` field inside the block content until one is found even if that `id` belongs to the an inner block, meaning the only way for block to return to null for `anchor` is for the content and all inner blocks to not have an anchor set.

The bigger problem is that this results in duplicate anchors. See screenshot below.
![image](https://user-images.githubusercontent.com/13604318/235830330-48bacb0e-dfc3-40e9-9328-2b00d4a7332b.png)

theodesp:
Hey @kidunot89  thanks for the report. Is it possible to use Anchors for Layout elements to avoid that issue in the meantime. I'm trying to think of a way to avoid searching the innerBlocks content for anchor ids.

Since we are rendering the whole block `render_block( $block )` and then using the following `$attribute_config`

```
$attribute_config = array(
									'type'      => 'string',
									'source'    => 'attribute',
									'attribute' => 'id',
									'selector'  => '*',
								);
```

It will search the innerHTML as well for that id. Not sure if there is an easy way to remove the innerHTML content.

theodesp:
Hey @kidunot89  I think I found the issue.

I can change this line:
https://github.com/wpengine/wp-graphql-content-blocks/blob/main/includes/Field/BlockSupports/Anchor.php#L40

into:

```
$rendered_block   = wp_unslash( $block['innerHTML'] );
```
**Before**
<img width="1537" alt="Screenshot 2023-05-03 at 12 30 54" src="https://user-images.githubusercontent.com/328805/235904238-ce1f1ab3-43a9-4232-9f62-368a3f80a5a1.png">

**After**
<img width="1508" alt="Screenshot 2023-05-03 at 12 29 04" src="https://user-images.githubusercontent.com/328805/235904055-ae3fcda7-fbec-4c6e-861b-92976be056b7.png">

theodesp:
https://github.com/wpengine/wp-graphql-content-blocks/pull/63

theodesp:
We will be releasing this tomorrow.

PR #63, opened 2023-05-03
Bug: inner blocks "anchor" field being applied to parent block resulting in duplicates

theodesp:
# Description

Bug fix for https://github.com/wpengine/wp-graphql-content-blocks/issues/62

Uses a special case for parsing `Anchor` fields. Instead of searching the whole block HTML we search for the first node id attribute.

### DIFF
--- a/includes/Field/BlockSupports/Anchor.php
+++ b/includes/Field/BlockSupports/Anchor.php
@@ -31,20 +31,12 @@ class Anchor {
 							'type'        => 'string',
 							'description' => __( 'The anchor field for the block.', 'wp-graphql-content-blocks' ),
 							'resolve'     => function ( $block ) {
-								$attribute_config = array(
-									'type'      => 'string',
-									'source'    => 'attribute',
-									'attribute' => 'id',
-									'selector'  => '*',
-								);
-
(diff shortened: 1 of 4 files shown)


includes/Field/BlockSupports/Anchor.php
```diff
@@ -31,18 +31,12 @@ public static function register( WP_Block_Type $block_spec ) {
...
-									'attribute' => 'id',
-									'selector'  => '*',
-								);
 								$rendered_block   = wp_unslash( render_block( $block ) );
 								$value            = null;
 								if ( empty( $rendered_block ) ) {
 									return $value;
 								}
-								$value = DOMHelpers::parseAttribute( $rendered_block, $attribute_config['selector'], $attribute_config['attribute'], null );
+								$value = DOMHelpers::parseFirstNodeAttribute( $rendered_block, 'id' );
```
theodesp, line 39: Special case for Anchor. We look only on the first node element for the attribute.

blakewilson (approved):
Well done @theodesp!

kidunot89:
@theodesp Phenomenal work 🚀 . I can confirm this solution works on my end as well.
image
Thanks for getting this patch so fasted.

PR #69, opened 2023-05-16
tests: fix Unit test autoloading and lint

justlevine:
This PR

1. Renames `WPGraphQL\ContentBlocks\Unit\RegistryTest` to `RegistryTestCase` to comply with PSR-4 autoloading.
2. Lints `DOMHelperTest` _using PHPCBF only_.

**Note:** The `composer.lock` file was _not_ regenerated, since I don't know what version of PHP ya'll are using to compile (to be addressed in a separate PR).

IDK what happened to the PR template, but 
- [x] I have signed the contributor agreement in the past

image

Issue #68, opened 2023-05-15
WPGraphQL Content Blocks Plugin Error


RohitDokhe:
Hello,

I have installed the “WPGraphQL Content Blocks” plugin and it activated successfully. However, I am getting an error when I try to open GraphQL IDE. It is showing a loader continues on screen and getting an error on screen “Parse error: syntax error, unexpected ‘WP_Block_Type’ (T_STRING), expecting function (T_FUNCTION) or const (T_CONST) in /var/www/html/wp-content/plugins/wp-graphql-content-blocks/includes/Blocks/Block.php on line 30”.

Here are all the details about my website: 
PHP: Version 7.3.17 
WordPress Version 6.2 
Theme: WordPress Twenty Twenty-Three Version 1.1 
Plugin: WPGraphQL Version 1.14.3, 
		WPGraphQL Content Blocks Version 0.2.1

Could you please help me resolve this issue?

justlevine:
> Here are all the details about my website:
> PHP: Version 7.3.17

@RohitDokhe The error you're seeing is due to the plugin's use of typed properties, which are only available in PHP 7.4. If you can update to that version (or even better PHP 8.0), that issue should go away.

PS @maintainers: running `composer phpcs` shows 17 PHP version compatibility issues and dozens of other failed sniffs.

![image](https://github.com/wpengine/wp-graphql-content-blocks/assets/29322304/483c9c6d-b543-4b71-8ec2-e12370570d63)

mindctrl:
We should probably go ahead and bump the min PHP version. We currently have min at 7.2 ([source](https://github.com/wpengine/wp-graphql-content-blocks/blob/56f1b1ec70e0478a768c65c47bfa609720863405/wp-graphql-content-blocks.php#L12)).

PR #70, opened 2023-05-16
chore: set Composer platform reqs

justlevine:
This PR sets the minimum PHP version in `composer.json`, to ensure that all local dependency builds are using the same version instead of the user's current PHP version.

This should help prevent PHP incompatibilities / breaking changes from going under the radar (e.g. #68), as well as ensure that any commits to `composer.lock` are valid regardless of the user's local env (e.g. #69 ).

- [x] Yes I signed the contributors agreement in the past

*Note*:
- Currently the minimum PHP requirement (`require->php` is set to 7.2 (per the readme.txt), despite the fact that the current codebase (per phpcs) is only compatible with PHP 7.4.
- The platform req (`config->platform->php`) however is set to *7.3*, as this is the minimum compatible version that `PHPUnit ^9.5` requires. This is _not_ ideal, and could cause breaking back-compat issues to sneak through in the future.
- Several composer dev-deps were (possibly) upgraded, due to the weak version constraints in `composer.json` and the requirement to run `composer update` to reset the packages to a PHP7.3 compatible version. Ideally those constraints would be more specific, but thats beyond the scope of this PR.

### DIFF
--- a/.changeset/two-numbers-end.md
+++ b/.changeset/two-numbers-end.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+chore: Set the minimum PHP version in `composer.json` to v7.2 (and the platform req to v7.3) to ensure contributions are built against the correct dependencies.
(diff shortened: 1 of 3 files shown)


theodesp (approved):

theodesp:
Ta!

PR #74, opened 2023-05-16
fix: [PHPCS] use `wp_rand()` instead of `rand()`

justlevine:
This PR replaces the usage of `rand()` with `wp_rand()` , per the project's PHPCS ruleset.

Since it's anyways being used for a rubbish name property, this should have no actual impact on usage.


theodesp (approved):

theodesp (approved):

theodesp:
Ta!

PR #76, opened 2023-05-16
dev: [PHPCS] rename `OnInit()` methods to `init()`

justlevine:
This PR renames the `WPGraphQL\ContentBlocks\Registry::OnInit()` and the `WPGraphQL\ContentBlocks\Type\Scalar::OnInit()` methods to `::init()` in order to comply with WPCS.

(Personally, I think the entire `Scalar` class should be renamed to `Scalar\BlockAttributesObject` and follow the established ecosystem pattern, but that's a (relatively) larger change with no immediate benefit).

PR #77, opened 2023-05-16
dev!: rename `BlockAttributesObject()` to `get_block_attributes_object_type_name()`

justlevine:
This PR renames the `WPGraphQL\ContentBlocks\Type\Scalar\Scalar::BlockAttributesObject()` method to the snake_cased `get_block_attributes_object_type_name()` to comply with WPCS while still remaining semantic.

**This is a breaking change.**

## Personal opinion
As mentioned in #76 , i'd much rather see *this PR rejected* and the codebase replace `Scalar` with:
```php
namespace WPGraphQL\ContentBlocks\Type\Scalar;
class BlockAttributesObject {
  public static function register();
  public static function get_type_name();
}
```
(since its the same level of 'breaking' for the api), but I'm trying to work within the codebase's existing patterns.

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -141,7 +141,7 @@ class Block {
 						break;
 					case 'array':
 					case 'object':
-						$graphql_type = Scalar::BlockAttributesObject();
+						$graphql_type = Scalar::get_block_attributes_object_type_name();
 						break;
 				}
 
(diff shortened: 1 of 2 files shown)


justlevine:
@theodesp is the changeset action semver compliant at _pre-release_? Wondering if I need to mark this as a `major` or a `minor`

theodesp (approved):

theodesp:
@justlevine everything is patch for now since the API is maturing.

theodesp (approved):

PR #78, opened 2023-05-16
dev: [PHPCS] deprecate unused `$context` param on `EditorBlockInterface::get_blocks()`

justlevine:
This PR deprecates the unused `$context` param from the `EditorBlockInterface::get_blocks()` method. All internal usage of that method has been updated as well.

Incidentally, the `EditorBlockInterfaceTest` class was also fixed to call that method statically instead of on an instance.

- [x] Yes I signed the contributor agreement

### DIFF
--- a/includes/Type/InterfaceType/EditorBlockInterface.php
+++ b/includes/Type/InterfaceType/EditorBlockInterface.php
@@ -25,7 +25,7 @@ final class EditorBlockInterface {
 	 *
 	 * @return \WP_Block_Type|null
 	 */
-	public static function get_block( array $block, #[Deprecated] $context = null ) {
+	public static function get_block( array $block, $context = null ) {
 		if ( null !== $context ) {
 			_deprecated_argument(
 				__METHOD__,
--- a/includes/Type/InterfaceType/EditorBlockInterface.php
+++ b/includes/Type/InterfaceType/EditorBlockInterface.php
@@ -2,7 +2,6 @@
 
 namespace WPGraphQL\ContentBlocks\Type\InterfaceType;
 
-use Exception;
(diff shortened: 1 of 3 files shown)


theodesp (approved):

theodesp (approved):

theodesp (approved):

theodesp (approved):

includes/Type/InterfaceType/EditorBlockInterface.php
```diff
@@ -21,12 +20,20 @@ final class EditorBlockInterface {
...
 	 *
-	 * @param array      $block   The block being resolved.
-	 * @param AppContext $context The AppContext.
+	 * @param array $block The block being resolved.
+	 * @param null  $context Deprecated.
 	 *
 	 * @return \WP_Block_Type|null
 	 */
-	public static function get_block( array $block, AppContext $context ) {
+	public static function get_block( array $block, #[Deprecated] $context = null ) {
```
mindctrl, line 28: Isn't `#[Deprecated]` a PHP 8 attribute? If so, the plugin header says we support 7.2.

justlevine: [I stand corrected](https://php.watch/versions/8.0/attributes#:~:text=The%20final%20syntax%20of%20Attributes,run%20on%20earlier%20PHP%20versions.), what I said above is only true if we make the params multiline.

Your call whether to break it multiline or just drop the 8.0 hinting for something that probably wont stay in the codebase for that long 🤷

mindctrl (approved):

theodesp (approved):

PR #79, opened 2023-05-17
dev!: rename `BlockAttributesObject()` to `get_block_attributes_object_type_name()` (#77)

theodesp:
#77

### DIFF
--- a/.changeset/rare-toes-brush.md
+++ b/.changeset/rare-toes-brush.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+Rename BlockAttributesObject() to get_block_attributes_object_type_name


.changeset/rare-toes-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
```
blakewilson, line 2: Wouldn't this be `minor` since it was breaking?

theodesp:
I don't this it's breaking. It doesn't change the response type. Only the function name:

theodesp:
> the name of a public static function changed, which means any references to the old method need to be replace with this one. And using this method is a best practice over hardcoding the name.

We don't support breaking changes on public since this plugin is an API and that would be very dangerous to do so since we can refactor the code and change some semantics at any time. As long as the API response stays the same then we should not mark it as breaking.

My safe assumption is that you should only rely on exposed `filters` and `actions` and maybe the Block base class here
https://github.com/wpengine/wp-graphql-content-blocks/blob/main/includes/Blocks/Block.php

But anything else is considered implementation detail that are undocumented and very risky to rely on.

justlevine:
Id argue that SemVer is about conveying _all breaking codebase changes_, not just changes that break the frontend schema. 
To me, your argument supports _being okay_ with releasing these sorts of 'technically-breaking but probably won't impact the user' more frequently than a breaking change that requires the user to rebuild their frontend, but that doesn't change the fact that the change needs to be communicated to people updating their plugins (especially if they're using composer).

But ultimately it's your project 🤷‍♂️

theodesp:
I think we need to specifically mention which classes are safe to extend and have the rest as unsafe and very risky to rely on.
Also anything that is undocumented is also risky.

theodesp:
Nice opinion post here:

https://gist.github.com/jashkenas/cbd2b088e20279ae2c8e

mindctrl:
Renaming a public function in a distributed plugin is a breaking change, even if it's not part of the public API we want people using. We just don't know how people have integrated with it. If we had a note in the phpdoc above the function or class that marked it as `@private` with a note explaining why, I think we could more comfortably deploy breaking changes like this without bigger version bumps.

Having said all that, I'm new on this project and don't have all the context yet. If we've communicated that this is pre-release software and breaking changes are likely, it's probably ok. But, looking at the README, I don't see any mention of instability. In [our docs](https://faustjs.org/docs/gutenberg/wp-graphql-content-blocks), we even say:

> This plugin is similar to the wp-graphql-gutenberg but it is actively maintained and tackles some scalability issues of the former.

I think this implies stability and not beta or pre-release. Unfortunate, as it does seem like we're still ramping up to a 1.0.

justlevine:
> Nice opinion post here:
> 
> https://gist.github.com/jashkenas/cbd2b088e20279ae2c8e

@theodesp Thanks for the link. 
Without weighing in on the opinion, I would like to point out that (as noted by many of the commenters) this only really applies to consumer software, not APIs (where the intended use _is_ for machine readability - i.e. composer - and semantic value to the user is just a potential bonus.

The entirety of the the plugin is an API, not just the generated schema additions (perhaps unlike other ecosystems, this is true for every single WPGraphQL extension ). The _intended_ use is for developers to implement and extend, so from a dependency management perspective, removing a public static PHP method needs to be treated as much as a first class citizen as removing a field from the schema 

> I think we need to specifically mention which classes are safe to extend and have the rest as unsafe and very risky to rely on.
Also anything that is undocumented is also risky.

I agree 💯. I also believe this has no direct bearing on whether something should be noted as a breaking change by a plugin that is currently using SemVer, in an ecosystem where everything else is using SemVer, and the recommended way of installation is with a SemVer package manager.

(As @mindctrl suggested, the exception should really only be a things that are explicitly documented as `@private` or `@internal` ).

Okay I've said my piece for real this time, and Ill go back to deferring to you ( didn't want to chance that whatever decision ya'll reach is based on a misunderstanding of the ecosystem and usage )

blakewilson:
I like the idea of using `@private` going forward. This will help us quantify if the changes we make are actually breaking or just our internal APIs.

blakewilson:
Or maybe even `@public` depending on the amount of APIs being public vs private.

mindctrl (approved):

PR #84, opened 2023-05-18
ci: wait until db is ready instead of sleeping fixed amount of time

mindctrl:
Removes the fixed 15 second sleep in favor of using `mysqladmin ping` to detect when the db is ready. Seems to save 5-6 secs per run in my tests, and in theory it would prevent failures if mysql took longer than 15 secs.


blakewilson (approved):
Nice! Looks like it's cut the wait time in half!

PR #85, opened 2023-05-18
chore: address phpcs issues

mindctrl:
Addresses remaining phpcs issues

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -223,8 +223,8 @@ class Block {
 					'name' => array(
 						'type'        => 'String',
 						'description' => __( 'The name of the block', 'wp-graphql-content-blocks' ),
-						'resolve'     => function ( $block, array $args, AppContext $context, ResolveInfo $info ) {
-							return $this->resolve( $block, $args, $context, $info );
+						'resolve'     => function ( $block ) {
+							return $this->resolve( $block );
 						},
 					),
 				),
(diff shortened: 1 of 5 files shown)


includes/Blocks/Block.php
```diff
@@ -235,12 +235,9 @@ private function register_type(): void {
...
 	 * Returns the necessary block data to resolve the block field.
 	 *
-	 * @param mixed                                $block   The block data passed to the resolver.
-	 * @param array                                $args    The arguments passed to the resolver.
-	 * @param \WPGraphQL\AppContext                $context The AppContext instance passed to the resolver.
-	 * @param \GraphQL\Type\Definition\ResolveInfo $info    The ResolveInfo instance passed to the resolver.
+	 * @param mixed $block The block data passed to the resolver.
 	 */
-	private function resolve( $block, array $args, AppContext $context, ResolveInfo $info ) {
+	private function resolve( $block ) {
```
justlevine, line 240: Premature optimization, but if you're changing the signature of this anyway, you might want to consider just changing the entire method to a callable, so you don't need to wrap it in an anonymous function for each `resolve` callback.

mindctrl: Thanks for the review! I'll log that for another PR. Trying to keep these coding standards PRs focused on the specific failures.

theodesp (approved):

PR #90, opened 2023-05-19
fix: [PHPStan] `$block_attributes` is always set

justlevine:
This PR fixes an issue in `Block::get_block_attributes_field()`, where the method was incorrectly checking if the paramater `$block_attributes` was passed to the method ( `isset()` ) instead of if the value passed was `null`.

The param signature has also been fixed and moved into PHP, to prevent errors like this from going unnoticed in the future.

Caught by PHPStan level 1.

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -144,11 +144,12 @@ class Block {
 	/**
 	 * Gets the WPGraphQL field registration config for the block attributes.
 	 *
-	 * @param array $block_attributes The block attributes.
+	 * @param ?array $block_attributes The block attributes.
 	 */
-	private function get_block_attribute_fields( $block_attributes ) {
+	private function get_block_attribute_fields( ?array $block_attributes ) {
 		$block_attribute_fields = array();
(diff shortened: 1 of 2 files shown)


includes/Blocks/Block.php
```diff
@@ -144,54 +144,62 @@ private function register_block_support_fields() {
...
+		// Bail early if no attributes are defined.
+		if ( null === $block_attributes ) {
+			return $block_attribute_fields;
+		}
+
+		foreach ( $block_attributes as $attribute_name => $attribute_config ) {
+			$graphql_type = null;
+
+			if ( ! isset( $attribute_config['type'] ) ) {
+				return $block_attribute_fields;
```
justlevine, line 161: Should this be returning instead of a `continue`?

Beyond the scope of this PR, but I didnt know a better way to flag it.

mindctrl (dismissed):

mindctrl (approved):

josephfusco (approved):
Thank you @justlevine!

PR #95, opened 2023-05-19
fix: cleanup unnecessary conditionals

justlevine:
This PR fixes issues with several conditional statements as detected by PHPStan level 4.


includes/Type/InterfaceType/EditorBlockInterface.php
```diff
@@ -26,6 +26,7 @@ final class EditorBlockInterface {
 	 * @return \WP_Block_Type|null
 	 */
 	public static function get_block( array $block, $context = null ) {
+		// @phpstan-ignore-next-line
```
justlevine, line 29: We have `$context` typed to `null` so linters and IDE can pick up that the parameter is deprecated. This should be removed along with the `$context` param in a breaking release.

theodesp (approved):

PR #101, opened 2023-05-22
fix: [PHPStan] better type checking in `ContentBlocksResolver::resolve_content_blocks()`

justlevine:
This PR fixes several issues in `ContentBlocksResolver::resolve_content_blocks()`, where a lack of type safety could cause a fatal PHP error.

Caught by PHPStan level 8

### DIFF
--- a/includes/Data/ContentBlocksResolver.php
+++ b/includes/Data/ContentBlocksResolver.php
@@ -32,13 +32,13 @@ final class ContentBlocksResolver {
 			// determining what fields should/should not be
 			// allowed to be returned?
 			$post    = get_post( $node->databaseId );
-			$content = $post->post_content;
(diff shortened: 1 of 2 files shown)


includes/Data/ContentBlocksResolver.php
```diff
@@ -32,13 +32,13 @@ public static function resolve_content_blocks( $node, $args, $allowed_block_name
 			// determining what fields should/should not be
 			// allowed to be returned?
 			$post    = get_post( $node->databaseId );
```
justlevine, line 34: I'm unclear why we're not just getting `$node->contentRaw`, but the existence of a comment block implies that its intentional, so I've left it alone.

mindctrl (approved):

PR #104, opened 2023-05-25
Add support for reusable blocks

williamjulianvicary:
This PR adds support for reusable blocks, rather than simply returning a core/block response. It detects if a reusable block has been found and if so, opens the post and parses the blocks from that, replacing the core/block with the newly parsed blocks.

This is probably a breaking change, given it's no longer sending a core/block response but it would equally be breaking if the core/block was mutated to embed the blocks within innerBlocks or similar. This in my opinion is the most compatible approach, it allows the reusable blocks to simply appear as if they were non-reusable from the GQL response perspective - allowing end users to change from reusable <> non-reusable with zero difference in the GQL response output - which feels sensible from my perspective.

### DIFF
--- a/.changeset/gorgeous-horses-arrive.md
+++ b/.changeset/gorgeous-horses-arrive.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": minor
+---
+
+Add support for Reusable Blocks
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ artifacts
 # docker
 .docker/plugins/
 !.docker/plugins/test-cpt
+docker-compose.override.yml
 
 # Built Zip
 @wpengine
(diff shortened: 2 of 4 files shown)


theodesp:
Hello @williamjulianvicary thanks for the contribution. Once we complete the CLA review process will will review the your PR!

theodesp:
Hey @williamjulianvicary Looks great. Could you also generate a changeset request as well?

You just need to run `npm run changeset` on the project folder and chose a `minor` change.

Also could you format the code and add a comment on the top? Something like:

```php
		// Resolve Reusable Blocks
		$new_parsed_blocks = array();
		foreach ( $parsed_blocks as $block ) {
			if ( $block['blockName'] === 'core/block' && $block['attrs']['ref'] ) {
				$reusable_blocks = parse_blocks( get_post( $block['attrs']['ref'] )->post_content );
				if ( ! empty( $reusable_blocks ) ) {
					array_push( $new_parsed_blocks, ...$reusable_blocks );
				}
			} else {
				$new_parsed_blocks[] = $block;
			}
		}

		$parsed_blocks = $new_parsed_blocks;
```

williamjulianvicary:
That's done

theodesp (approved):
LGTM!

PR #105, opened 2023-05-25
Fix regression - Allow blocks with names to pass filtering, even if empty when rendered

williamjulianvicary:
Adds a fix for a regression caused by #59 

Blocks may intentionally be empty rendered, and the properties passed to the frontend for rendering on the frontend the previous change in this file caused a regression which blocked those use cases, while adding support for Classic Editor blocks (which didn't have a name and were therefore filtered).

This PR allows both:
- If blocks have a name, they pass the filter regardless.
- Then the block is rendered and checked for empty()

As a side-effect, this should be more performant as it's avoiding a render.

### DIFF
--- a/includes/Data/ContentBlocksResolver.php
+++ b/includes/Data/ContentBlocksResolver.php
@@ -57,6 +57,10 @@ final class ContentBlocksResolver {
 		$parsed_blocks = array_filter(
 			$parsed_blocks,
 			function ( $parsed_block ) {
+				if (! empty( $parsed_block['blockName'] )) {
+					return true;
+				}
+
 				// Strip empty comments and spaces
 				$stripped = preg_replace( '/<!--(.*)-->/Uis', '', render_block( $parsed_block ) );
 				return ! empty( trim( $stripped ?? '' ) );


theodesp:
Hey @williamjulianvicary since we allow empty blocks, should we just remove the whole filtering logic altogether?

```php
// 1st Level filtering of blocks that are empty
		$parsed_blocks = array_filter(
			$parsed_blocks,
			function ( $parsed_block ) {
				// Strip empty comments and spaces
				$stripped = preg_replace( '/<!--(.*)-->/Uis', '', render_block( $parsed_block ) );
				return ! empty( trim( $stripped ?? '' ) );
			},
			ARRAY_FILTER_USE_BOTH
		);
```

williamjulianvicary:
I don't quite understand the implications of that yet, but I think this is tied to classic blocks. In the tests there are fake posts with content like:

```
  <!--  -->
                <!--  -->
                <!-- wp -->
                <!-- /wp -->

                <!-- wp: -->
                <!-- /wp: -->
```

My assumption was that the change was to support keeping classic blocks (which are not empty, but do not have a name), still strip the above empty-non-named blocks out of the CSS and my addition then allows blocks that are named but also empty.

Note: The logic I've added, is actually reverting part of the change that was added in the recent addition of Classic Block support, originally it only checked the name - this didn't work for Classic Editor blocks, but the change then caused a regression with named empty blocks - my change here is retaining the new logic for removing empty blocks but adding back in the retention of named blocks.

Make sense? TL;DR I think both pieces of logic still have a place, removal will break tests, and I'm not sure on the knock on effect of retaining those empty blocks on the graphql output without running some tests.

theodesp (approved):
LGTM!

PR #106, opened 2023-05-25
Only register the anchor once, then apply to the block

### LINKED ISSUE #98: Bug: `BlockWithSupportsAnchor` is causing `DUPLICATE_TYPE` error
When querying for editor blocks with `GRAPHQL_DEBUG` enabled, I get the following `DUPLICATE_TYPE` error:
`You cannot register duplicate Types to the Schema. The Type 'BlockWithSupportsAnchor' already exists in the Schema. Make sure to give new Types a unique name.`

This is due to the fact that that `register_graphql_interface()` is being called [for every block type](https://github.com/wpengine/wp-graphql-content-blocks/blob/main/includes/Blocks/Block.php#L141). 

What should actually happen is that the `BlockWithSupportsAnchor` should be registered once in the `Registry`, and `Block::register_block_supports_field()` should only `register_interface_to_type()`. 

## Example

image

williamjulianvicary:
Fixes #98.

The interface was being registered for any block that supported anchors, if more than one block supported them then GraphQL was complaining about duplicate entries. This moves the registering of the interface to the registry and then the application of that interface within the block registering flow only registers that interface to the type/block.

### DIFF
--- a/.changeset/great-ravens-joke.md
+++ b/.changeset/great-ravens-joke.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": minor
+---
+
+Fix regression with addition of anchor support - only register interface once
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -135,7 +135,7 @@ class Block {
 	 * Registers fields for the block supports.
 	 */
 	private function register_block_support_fields(): void {
-		Anchor::register( $this->block );
+		Anchor::register_to_block( $this->block );
 	}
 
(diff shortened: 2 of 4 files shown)


theodesp:
Thank you @williamjulianvicary  will review

theodesp:
Hey @williamjulianvicary Looks great. Could you also generate a changeset request as well?

You just need to run npm run changeset on the project folder and chose a minor change.

williamjulianvicary:
That's done!

theodesp (approved):
LGTM!

PR #110, opened 2023-06-07
chore: add roave/security-advisories Composer dependency

mindctrl:
Adds the `roave/security-advisories` dependency to help ensure we don't install dependencies with known vulnerabilities.

We have this in Faust too ([reference](https://github.com/wpengine/faustjs/blob/4cae3d9fc449bb52beb30cdedfb311806d9b2777/plugins/faustwp/composer.json#LL22C5-L22C47)).

[More info](https://github.com/Roave/SecurityAdvisories).


theodesp (approved):
LGTM!

PR #114, opened 2023-06-12
Chore: CoreSeparator Changeset

theodesp:
# Description
Adds cssClassName attribute in CoreSeparator

#Test
Check that the cssClassName attribute is available in the CoreSeparator block.

### DIFF
--- a/.changeset/friendly-horses-pay.md
+++ b/.changeset/friendly-horses-pay.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+Chore: Changeset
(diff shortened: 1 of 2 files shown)


includes/Blocks/CoreSeparator.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
mindctrl, line 1: @theodesp not sure why this file is showing, since the same was merged in #113, unless you branched without updating `main` first?

mindctrl (approved):

.changeset/friendly-horses-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+Chore: Changeset
```
blakewilson, line 5: @theodesp We should probably make this more descriptive as it doesn't describe the changes related to `CoreSeparator`

PR #120, opened 2023-06-21
[Merl-786] `post_type_supports` -> `use_block_editor_for_post_type`

### LINKED ISSUE #27: blocks are being exposed to all post types, even those that don't use Gutenberg
NOTE: This was reported to me in Slack by @kidunot89, just passing it on here: 

> Might want to tell to Content Blocks team to adjust the way they are targeting CPTs, they are adding editorBlocks to all of them even the ones that don't use gutenberg.

image

theodesp:
# Description

Replaces the `post_type_supports` with `use_block_editor_for_post_type` when resolving the `get_supported_post_types`. This has the added benefit of running the [use_block_editor_for_post_type](https://developer.wordpress.org/reference/hooks/use_block_editor_for_post_type/) hook that was missed, potentially including blocks that dont work with the block editor. 

## Tasks
- [x] Has automated tests confirming schema accuracy on post types that support blocks and post types that do not support blocks.
- [x] Has been tested with [WPGraphQL for WooCommerce](https://github.com/wp-graphql/wp-graphql-woocommerce/releases)
- [x] Has been tested with WPGraphQL for ACF ([existing](https://github.com/wp-graphql/wp-graphql-acf) and [new](https://github.com/wp-graphql/wpgraphql-acf))
- [x] Has a changeset

Closes #27

### DIFF
--- a/.changeset/odd-files-attend.md
+++ b/.changeset/odd-files-attend.md
@@ -1,5 +1,6 @@
 ---
-"@wpengine/wp-graphql-content-blocks": patch
+"@wpengine/wp-graphql-content-blocks": major
 ---
 
 Fix: use `use_block_editor_for_post_type` instead of `post_type_supports` when filtering the post types.
+WARNING: Potential BREAKING schema changes on previously exposed blocks that do not support the block editor. Those blocks will no longer inherit the `editorBlocks` field.
--- a/.changeset/odd-files-attend.md
+++ b/.changeset/odd-files-attend.md
@@ -3,4 +3,4 @@
 ---
 
 Fix: use `use_block_editor_for_post_type` instead of `post_type_supports` when filtering the post types.
-WARNING: Potential BREAKING schema changes on previously exposed blocks that do not support the block editor. Those blocks will no longer inherit the `editorBlocks` field.
+**BREAKING**: Potential schema changes on previously exposed blocks that do not support the block editor. Those blocks will no longer inherit the `editorBlocks` field.
--- a/.changeset/odd-files-attend.md
+++ b/.changeset/odd-files-attend.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+Fix: use `use_block_editor_for_post_type` instead of `post_type_supports` when filtering the post types.
--- a/includes/Utilities/WPHelpers.php
+++ b/includes/Utilities/WPHelpers.php
@@ -33,11 +33,10 @@ final class WPHelpers {
 		if ( empty( $block_editor_post_types ) || ! is_array( $block_editor_post_types ) ) {
 			return $supported_post_types;
 		}
-
 		// Iterate over the post types
 		foreach ( $block_editor_post_types as $block_editor_post_type ) {
 			// If the post type doesn't support the editor, it's not block-editor enabled
-			if ( ! post_type_supports( $block_editor_post_type->name, 'editor' ) ) {
+			if ( ! use_block_editor_for_post_type( $block_editor_post_type->name ) ) {
 				continue;
 			}
 
--- a/tests/unit/RegistryTestCase.php
+++ b/tests/unit/RegistryTestCase.php
@@ -65,6 +65,42 @@ final class RegistryTestCase extends PluginTestCase {
 	}
 
 	/**
+	 * This test ensures that when disabling the block editor for post types then
+	 * no additional interfaces are included in them.
+	 */
+	public function test_no_additional_interfaces_on_block_editor_disabled_block_types() {
+		add_filter('use_block_editor_for_post_type', '__return_false');
+		$query = '
+		query GetType($name:String!) {
+			__type(name: $name) {
+				interfaces {
+					name
+				}
+			}
+		}
+		';
+
+		$this->instance->init();
+
+		// Verify the response contains what we put in cache
+		$response           = graphql(
+			array(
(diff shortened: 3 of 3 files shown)


mindctrl:
Let's get some tests that confirm the schema is correct for post types that do and do not support blocks.

jasonbahl:
@theodesp any updates on when this might be released? 

I'm trying to use WPGraphQL Content Blocks + ACF Blocks, and anytime I register a block to be available on only one specific post type, I am seeing it exposed to the Schema on _all_ post types, which is not ideal.

jasonbahl:
> I'm trying to use WPGraphQL Content Blocks + ACF Blocks, and anytime I register a block to be available on only one specific post type, I am seeing it exposed to the Schema on all post types, which is not ideal.

Ah, I think I'm running into a different, but related issue. 

If I register a a block and set it to only be exposed to a single post type, it still implements the interfaces for _all_ post types.

For example, I've registered a block that is only available on the "Post" post type.

But it gets the following interfaces applied to it:

CleanShot 2023-08-17 at 10 57 51

And it can be queried on _any_ post type, not _just_ the Post Post Type.

I would expect a Block registered to just one Post Type to have the Interface for that Post Type implemented. In my case, I would expect `PostEditorBlock` to be implemented but not `EventEditorBlock`, `PageEditorBlock`, etc.

jasonbahl:
Update: 

I think what I'm doing is a bit custom and not necessarily expected to be working. 

I'm using ACF Extended which provides a way to register block types and constrain which post types the block type should be available on. 

The Blocks registered with constraints are still being exposed in WPGraphQL Content Blocks to _all_ post types via `Registry->get_block_context_interfaces( $block_name )`

This is because ACF Extended adds a `post_types => []` configuration to each block type, but this isn't actually something supported by core Gutenberg (yet, at least as far as I can tell: https://github.com/WordPress/gutenberg/issues/9855#issue-359736695)

So, because this is added functionality of ACF Extended, there's no reasonable expectation WPGraphQL Content Blocks would have supported this. 

I'll open a new issue and corresponding PR to add a filter so 3rd parties can hook in and customize some of the logic. 

I'm supporting ACF Extended Fields in WPGraphQL for ACF and since I'm also working on supporting ACF Blocks + WPGraphQL Content Blocks in that codebase, I would like to filter this logic so things play nice, but I'll need a filter to exist.

jasonbahl:
I opened an issue (#147) and PR (#148) to help with the situation I was facing. 🙏🏻

theodesp:
Check with with [WPGraphQL for WooCommerce](https://github.com/wp-graphql/wp-graphql-woocommerce/releases)

Created one product:

Screenshot 2023-10-02 at 13 45 27

theodesp:
Check with with [wp-graphql-acf] old and new.

Created one block:
Screenshot 2023-10-02 at 14 39 09

theodesp:
Using `add_filter('use_block_editor_for_post_type', '__return_false');`

Screenshot 2023-10-04 at 11 44 56
 the editorBlocks is not registered

mindctrl (commented):
This LGTM. Would be nice to have tests that confirm the desired behavior.

jasonbahl:
One thing to note: 

This will cause breaking changes to the Schema on any site that updates.

.changeset/odd-files-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
```
mindctrl, line 2: As @jasonbahl noted, this will change the schema in a breaking way. Should we make this a major bump? Thoughts @theodesp @blakewilson?

theodesp: I would argue that this fixes a bug instead of introducing a feature.

blakewilson: @theodesp Does fixing the bug cause breaking changes though? I'm thinking about it from a user's perspective: If they see a `patch` release for a bug fix, download the latest, and their environment breaks, a `major` bump may be more appropriate here.

justlevine: >  If they see a patch release for a bug fix, download the latest, and their environment breaks, a major bump may be more appropriate here.

Worse, even if they have major autoupdates disabled or are using semver in Composer, then this break can happen automatically without anyone needing to actively "download the latest".

mindctrl: I think we should do the major bump. The break is there regardless. SemVer says this about bug fixes:

> Bug fixes not affecting the API increment the patch version, backward compatible API additions/changes increment the minor version, and backward incompatible API changes increment the major version.

.changeset/odd-files-attend.md
```diff
@@ -0,0 +1,6 @@
+---
+"@wpengine/wp-graphql-content-blocks": major
+---
+
+Fix: use `use_block_editor_for_post_type` instead of `post_type_supports` when filtering the post types.
+WARNING: Potential BREAKING schema changes on previously exposed blocks that do not support the block editor. Those blocks will no longer inherit the `editorBlocks` field.
```
blakewilson, line 6:
```suggestion
**BREAKING**: Potential schema changes on previously exposed blocks that do not support the block editor. Those blocks will no longer inherit the `editorBlocks` field.
```

Small nitpick, but let's call out "breaking" at the beginning of the line in the case someone quickly skims the release notes before updating.

blakewilson (approved):

PR #123, opened 2023-06-26
Add `cssClassName` to CoreList block

blakewilson:
Adds the `cssClassName` attribute to the `CoreList` block similar to https://github.com/wpengine/wp-graphql-content-blocks/pull/122

## Testing

1. Create a `CoreList` block on a page or post
2. Query for that page/post in WPGraphQL GraphiQL
3. Query the `cssClassName` attribute on the `CoreList` block
4. Notice the class names come back as expected

Screenshot 2023-06-26 at 11 39 23 AM

### DIFF
--- a/.changeset/new-planes-itch.md
+++ b/.changeset/new-planes-itch.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": minor
+---
+
+Adds the `cssClassName` attribute to the `CoreList` block. This allows you to query for the proper class names that WordPress assigns to the Core List block.
--- a/includes/Blocks/CoreList.php
+++ b/includes/Blocks/CoreList.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Core List Block
+ *
+ * @package WPGraphQL\ContentBlocks\Blocks
+ */
+
+namespace WPGraphQL\ContentBlocks\Blocks;
+
+/**
+ * Class CoreList
+ */
+class CoreList extends Block {
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @var array|null
+	 */
+	protected ?array $additional_block_attributes = array(
+		'cssClassName' => array(
+			'type'      => 'string',
+			'selector'  => 'ul,ol',
+			'source'    => 'attribute',
+			'attribute' => 'class',
+		),
+	);
+}


theodesp (approved):

mindctrl (approved):

jasonbahl:
Also what's the difference between `className` and `cssClassName`?

blakewilson:
@jasonbahl The `cssClassName` attribute that is returned is a string with each class name space separated i.e. `has-background-color has-text-color has-background has-large-font-size` so `String` should be the correct type here.

blakewilson:
> Also what's the difference between `className` and `cssClassName`?

I'll refer to @theodesp for this one. He can answer it better than I can.

jasonbahl:
@blakewilson it looks like Gutenberg defines the list block to NOT have cssClassName support: https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks/list/block.json#L40

This seems intentional from the Gutenberg core team, and if WPGraphQL Content Blocks isn't outputting it, this seems like correct behavior.

theodesp:
We do have a relevant ticket to generate those cssClassNames in Faust  using a helper function instead of asking the API. Until we implement this, we are just extracting it from the HTML. 
```
impor {getClasses} from '@faustwp/blocks'
const classNames = getClasses(theme, props)
```
However there are a lots of edge cases to consider here so the most obvious way is to ask the API for those.

Ideally once done, we will deprecate the `cssClassName` attributes for the blocks.

justlevine:
> @blakewilson it looks like Gutenberg defines the list block to NOT have cssClassName support: https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks/list/block.json#L40
> 
> This seems intentional from the Gutenberg core team, and if WPGraphQL Content Blocks isn't outputting it, this seems like correct behavior.

@jasonbahl `supports.className` in the .json refers to [the addition of a `.my-block-class`](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#classname) to the component, not whether component behavior is determined by CSS classes.  The use of GB utility classes and design tokens are determined by the individual controls and that's what's being returned by `cssClassName`.

(Personally agree the `String` return type is correct in for this use case, and really it's the field name that should be semantically changed to `cssClasses`, but that would be a breaking schema change, and it probably doesn't deserve a v2.0 if the goal is to deprecate the field altogether)

theodesp:
> > @blakewilson it looks like Gutenberg defines the list block to NOT have cssClassName support: https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks/list/block.json#L40
> > This seems intentional from the Gutenberg core team, and if WPGraphQL Content Blocks isn't outputting it, this seems like correct behavior.
> 
> @jasonbahl `supports.className` in the .json refers to [the addition of a `.my-block-class`](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#classname) to the component, not whether component behavior is determined by CSS classes. The use of GB utility classes and design tokens are determined by the individual controls and that's what's being extracted from the generated markup and returned by `cssClassName`.
> 
> (Personally agree the `String` return type is correct in for this use case, and really it's the field name that should be semantically changed to `cssClasses`, but that would be a breaking schema change, and it probably doesn't deserve a v2.0 if the goal is to deprecate the field altogether)

Yes. This should have been a property of each block. Unfortunately the exposed `className` only refers to the additional className you can add on each block and not the whole list of classNames.

blakewilson:
Thanks for the context @theodesp, and thanks @jasonbahl @justlevine for your input here. This'll be a good PR to reference for historical purposes around this property.

Issue #91, opened 2023-05-19
Consider aligning PHPCS ruleset with WPGraphQL.


justlevine:
In order to make it easier for other developers in the ecosystem to contribute to this plugin,would it be possible to align with some of the [ruleset options used by WPGraphQL core](https://github.com/wp-graphql/wp-graphql/blob/develop/phpcs.xml.dist) - or at least the ones that dont conflict too bad with the included WPEngine ruleset (I'm not particularly familar with what ya'll usually use internally)?

(I'd be particularly nice if in 2023 we could be all be using short array syntax).

Thanks for considering!

justlevine:
To further this goal across the ecosystem (and a deeper explanation on why a shared ruleset is important), I released this [WPGraphQL Coding Standards](https://github.com/axewp/WPGraphQL-Coding-Standards), which should make it significantly easier to evaluate the impact and value on this codebase.

To test:
1. run `composer require --dev axepress/wp-graphql-cs` locally to install the sniffs.
2. Copy the example `phpcs.xml.dist.example` file over to the root directory and rename it to `phpcs.xml`. 
3. Replace the included/excluded directories in `phpcs.xml` with those from `phpcs.xml.dist`
4. Run `vendor/bin/phpcs` and review the results (or even better run `vendor/bin/phpcbf`, and then diff the branch).

theodesp:
@justlevine Thanks. We do maintain a phpcs plugin

https://github.com/wpengine/wpengine-coding-standards

do you think we can include this ruleset there?

justlevine:
@theodesp looking at the ruleset, the sniffs WPEngine-Strict has cherry-picked are already included in `WPGraphQL-Minimum`, as are all the ones in the `phpcs.xml.dist` of this repo.

The only conflict I see is the wp-core preference for long `array()` syntax - WPGraphQL core (and the ruleset) enforces the short `[]` syntax.

The next level up, `WPGraphQL-Strict` includes functional sniffs about PHP best practices, so that also should be fine.

[...]

PR #126, opened 2023-07-05
POC: Use WPGraphQL Coding Standards

justlevine:
This PR illustrates the effects of bringing this projects coding standards in-line with those used by WPGraphQL, as proposed in #91 . It is _not_ meant to be merged.

The `wp-graphql-cs` library is just included for ease of example, and is not a requisite for adopting the rules it contains.

`WPGraphQL-Minimum` applies the rules currently used by WPGraphQL core. 
The `WPGraphQL-Strict` ruleset is a proposal for core and includes a significant number of functional sniffs. See: https://github.com/wp-graphql/wp-graphql/pull/2830
The `WPGraphQL` ruleset includes docs and formatting sniffs.

Note: The wpengine-cs library was removed, as it hasnt been updated in 3 years, and the dependencies were conflicting. The only explicit rules from `WPEngine-Strict` that arent already included in the ecosystem ruleset are included in `.phpcs.xml.dist`

Best way to review would be to traverse the diffs of the individual commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,5 +23,8 @@ artifacts
 !.docker/plugins/test-cpt
 docker-compose.override.yml
 
+# phpcs cache
+/tests/_output/cache.json
+
 # Built Zip
 @wpengine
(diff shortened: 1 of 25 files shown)


mindctrl:
@justlevine thank you for this POC. I think these are good suggestions. I'll get this on our todo list and revisit soon.

justlevine:
Superseded by #162

PR #130, opened 2023-07-11
Bug: CoreImage width attribute resolve throws error.

theodesp:
# Description
Fixed bug when CoreImage width attribute fails to resolve thowing error.

# Testing
1 - Perform a query on the width attributes of the `CoreImage` and `CoreColumn` Blocks using fragments. You can look at the test case query for example.

2 - Query fails with the following error:

Fields \"attributes\" conflict because subfields \"width\" conflict because they return conflicting types Float and String.

Afte the fix this should resolve without issues.

Screenshot 2023-07-11 at 12 32 27

### DIFF
--- a/.changeset/serious-items-bow.md
+++ b/.changeset/serious-items-bow.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+Bug Fix: CoreImage `width` attribute throws error.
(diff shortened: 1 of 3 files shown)


mindctrl (commented):
What tests could we write to ensure we don't have this bug again?

theodesp:
> What tests could we write to ensure we don't have this bug again?

Added a test case.

mindctrl (approved):

Issue #135, opened 2023-07-24
CoreVideo returns false for controls regardless of admin setting


DesignyourCode:
When fetching the CoreVideo settings, it returns false regardless of what has been set in the admin.

I have tried to save it when set to true, and false then true and it always comes through as false.

Removing and re-adding the component also doesn't help.

theodesp:
Thanks for the report. Will try to reproduce.

theodesp:
Hey @DesignyourCode  I was able to reproduce. I will log a ticket and provide a fix soon. Thank you!

theodesp:
@theodesp you could temporarily fix this by adding the following piece of code next to the following line. (although we will try to refactor the code a bit in the fix).

https://github.com/wpengine/wp-graphql-content-blocks/blob/main/includes/Blocks/Block.php#L293

```
if ( 'boolean' === $attribute_config['type'] ) {
				return ! is_null( $value );
			}
```

PR #138, opened 2023-07-26
Bug Fix: Parsing boolean block attributes like in CoreVideo always return false.

theodesp:
# Description

Bug fix. Parsing boolean block attributes like in CoreVideo always return false. 

Fixes: https://github.com/wpengine/wp-graphql-content-blocks/issues/135

# Testing

1. Add a CoreVideo block in a Post and enable some of its checkbox settings for autoplay on the side.
2. Try to query the block attributes.
3. The boolean attributes should return the correct checked/unchecked value.

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -286,10 +286,15 @@ class Block {
 
 					break;
 			}//end switch
-
-			// if type is set to integer, get the integer value of the attribute.
-			if ( 'integer' === $attribute_config['type'] ) {
-				$value = intval( $value );
+			// Post processing of return value based on configured type
+			switch ( $attribute_config['type'] ) {
+				case 'integer':
+					$value = intval( $value );
+					break;
+				case 'boolean':
+					// Only false when value is not null or actually false
+					$value = ! ( false === $value || is_null( $value ) );
+					break;
 			}
 
 			return $value;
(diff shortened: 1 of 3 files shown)


mindctrl (changes requested):
LGTM. One suggestion to clarify the changelog entry.

mindctrl:
I fixed the Sonarqube issue. It was complaining about the `http` URLs in the tests.

includes/Blocks/Block.php
```diff
@@ -286,10 +286,15 @@ private function resolve_block_attributes( $block, $attribute_name, $attribute_c
...
-			if ( 'integer' === $attribute_config['type'] ) {
-				$value = intval( $value );
+			// Post processing of return value based on configured type
+			switch ( $attribute_config['type'] ) {
+				case 'integer':
+					$value = intval( $value );
+					break;
+				case 'boolean':
+					// Only false when value is not null or actually false
+					$value = ! ( false === $value || is_null( $value ) );
```
justlevine, line 296: Do we need to be concerned with other falsey values like an empty string?

Im not too familiar with the Block Attribute [type validation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-attributes/#type-validation), but WP is usually horrible at strict booleans.

mindctrl (dismissed):

includes/Blocks/Block.php
theodesp: I updated the code to handle null returns now. This should work as expected.

matthewguywright (approved):

PR #155, opened 2023-10-09
CI/CD: Make release dependent on tests.

theodesp:
# Description

* Runs Release step when both "Test Plugin" and "End-to-End Tests" are successful.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,13 +1,16 @@
 name: Release
 
 on:
-  push:
-    branches:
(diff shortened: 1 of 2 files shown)


TeresaGobble:
Thank you for this addition, @theodesp! This is going to be such a big safety feature for releases. 🚀 

Would you please provide some instruction on how to test this out? Testing for release behavior is a bit tricky, and I'm not sure what the best workflow is for it.

blakewilson (approved):

PR #162, opened 2023-11-27
Chore/wp graphql cs

theodesp:
# Description

Merge of https://github.com/wpengine/wp-graphql-content-blocks/pull/126 with some ignores. Commit history is maintained.

# Testing

1. Setup docker-compose
2. Run unit tests
3. Run lint tests

They should all pass

PR #163, opened 2023-11-27
chore: upgrade wp-graphql-cs to `2.0.0-beta`

justlevine:
This PR updates the plugin to use `axepress/wp-graphql-cs:2.0.0-beta`, which is based on WPCS 3.0+.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -49,6 +49,11 @@
 		<exclude name="Squiz.Commenting.InlineComment.InvalidEndChar" />
 		<!-- This should be excluded upstream in the ruleset-->
 		<exclude name="WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase" />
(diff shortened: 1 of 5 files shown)


theodesp:
Thanks since its a branch of a branch lets review and merge the first branch and then review this branch ;P

justlevine:
> Thanks since its a branch of a branch lets review and merge the first branch and then review this branch ;P

Your call. Did it this way instead of directly as suggestions on #162 because it was faster for me and didn't seem worthy of a separate changeset item, but your call if you prefer to wait until that's merged and then change the base branch here 🤷

mindctrl (approved):

justlevine:
@theodesp was this closed intentionally, or a side effect of deleting the target branch?

theodesp:
@justlevine probably a side-effect since this action happened automatically. Could you re-submit on a new branch based on the current main please?

Issue #61, opened 2023-05-02
Add support for extracting array type query data from blocks


theodesp:
# Description
Certain types of blocks have an `array` type of attribute which means that there is a nested object types that we need to extract.
Currently we do not support those so we want to provide this functionality as well.

For example:

* [CoreGallery](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/gallery/block.json#L12)
* [CoreTable](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/table/block.json#L25-L64)

# Technical notes

* The code to detect the attribute type [here](https://github.com/wpengine/wp-graphql-content-blocks/blob/main/includes/Blocks/Block.php#L129-L146) should also handle the `array` type.
* This should be a recursive step since it can have many levels of inner attributes.

![image](https://user-images.githubusercontent.com/328805/235649888-6f641bec-3a6d-47b7-86f5-03f61c24f54d.png)

PR #176, opened 2024-01-02
Add support for querying array type query data from blocks

theodesp:
# Description

Fixes: https://github.com/wpengine/wp-graphql-content-blocks/issues/61

# How to Test

* Create a Table Block and add a header footer and caption
* Add also some cells
* You will be able to query the head, foot and content cells

# Screenshots

Screenshot 2024-01-02 at 11 31 02

### DIFF
--- a/.changeset/ten-dolphins-smash.md
+++ b/.changeset/ten-dolphins-smash.md
@@ -1,5 +1,5 @@
 ---
-"@wpengine/wp-graphql-content-blocks": minor
+"@wpengine/wp-graphql-content-blocks": major
 ---
 
 Feature: Add support for querying array type query data from blocks
--- a/.changeset/ten-dolphins-smash.md
+++ b/.changeset/ten-dolphins-smash.md
@@ -1,0 +1,7 @@
+---
+"@wpengine/wp-graphql-content-blocks": minor
+---
+
+Feature: Add support for querying array type query data from blocks
+
+Query source block attribute types are supported. See: https://developer.wordpress.org/block-editor/reference-guides/block-api/block-attributes/#query-source
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -81,13 +81,14 @@ class Block {
 	 * Registers the block attributes GraphQL type and adds it as a field on the Block.
 	 */
 	private function register_block_attributes_as_fields(): void {
+		// Grab any additional block attributes attached into the class itself
 		if ( isset( $this->additional_block_attributes ) ) {
-			$block_attributes       = ! empty( $this->block_attributes ) ? array_merge( $this->block_attributes, $this->additional_block_attributes ) : $this->additional_block_attributes;
-			$block_attribute_fields = $this->get_block_attribute_fields( $block_attributes );
+			$block_attributes = ! empty( $this->block_attributes ) ? array_merge( $this->block_attributes, $this->additional_block_attributes ) : $this->additional_block_attributes;
 		} else {
-			$block_attribute_fields = $this->get_block_attribute_fields( $this->block_attributes );
+			$block_attributes = $this->block_attributes;
 		}
-
+		$block_attribute_fields = $this->get_block_attribute_fields( $block_attributes, $this->type_name . 'Attributes' );
+		// For each attribute, register a new object type and attach it to the block type as a field
 		if ( ! empty( $block_attribute_fields ) ) {
 			$block_attribute_type_name = $this->type_name . 'Attributes';
 			register_graphql_object_type(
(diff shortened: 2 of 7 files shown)


tests/unit/blocks/CoreVideoTest.php
```diff
@@ -80,16 +80,16 @@ public function test_retrieve_core_video_attributes() {
...
-            'muted' => false,
-            'caption' => '',
+            'tracks' => [],
+            'muted' => null,
+            'caption' => null,
             'preload' => 'auto',
             'src' => 'http://mysite.local/wp-content/uploads/2023/07/pexels_videos_1860684-1440p.mp4',
             'playsInline' => true,
-            'controls' => null,
+            'controls' => true,
```
theodesp, line 89: Default value of `controls` is `true`

tests/unit/blocks/CoreVideoTest.php
```diff
@@ -80,16 +80,16 @@ public function test_retrieve_core_video_attributes() {
             'align' => null,
             'anchor' => null,
             'autoplay' => true,
-            'tracks' => '[]',
-            'muted' => false,
-            'caption' => '',
+            'tracks' => [],
+            'muted' => null,
+            'caption' => null,
```
theodesp, line 85: caption is not set so its `null`

justlevine (commented):
For this new set of private methods, can we put the param/return types _inside_ the function signatures instead of as doc-types?

Stricter types are always better in terms of tech-debt and extensibility.

includes/Blocks/Block.php
```diff
@@ -122,65 +123,184 @@ private function register_block_attributes_as_fields(): void {
...
-			return $block_attribute_fields;
-		}
-
-		foreach ( $block_attributes as $attribute_name => $attribute_config ) {
-			$graphql_type = null;
-
-			if ( ! isset( $attribute_config['type'] ) ) {
-				return $block_attribute_fields;
-			}
+	private function get_attribute_type( $name, $attribute, $prefix ): mixed {
```
mindctrl, line 132: I believe `mixed` is only available in PHP 8.0+. The plugin currently works with PHP 7.4, so this would fatal error on sites using that. Not opposed to bumping min requirements of the plugin, but that's something we should discuss first.

includes/Blocks/Block.php
```diff
@@ -237,72 +357,93 @@ private function resolve( $block ) {
...
+		return $result;
+	}
 
+	/**
+	 * Parses the block content of a source only block type
+	 * 
+	 * @param string $html The html value
+	 * @param string $source The source type
+	 */
+	private function parse_single_source( $html, $source ): string|null {
```
mindctrl, line 437: Union types like `string|null` are not compat with PHP 7.4.

includes/Blocks/Block.php
```diff
@@ -122,65 +123,184 @@ private function register_block_attributes_as_fields(): void {
...
+
+	/**
+	 * Gets the WPGraphQL field registration config for the block attributes.
+	 *
+	 * @param ?array $block_attributes The block attributes.
+	 * @param string $prefix The current prefix string to use for the get_query_type
+	 * 
+	 * @return array
+	 */
+	private function get_block_attribute_fields( ?array $block_attributes, $prefix = null ): array {
```
mindctrl, line 191: `$prefix = null` doesn't match the phpdoc type of `string`. I'd recommend `$prefix = ''` here to match.

includes/Blocks/Block.php
```diff
@@ -122,65 +123,184 @@ private function register_block_attributes_as_fields(): void {
...
+	}
+	
+	/**
+	 * Returns the type of the block query attribute
+	 * 
+	 * @param string $name The block name
+	 * @param array  $query The block query config
+	 * @param string $prefix The current prefix string to use for registering the new query attribute type
+	 */
+	private function get_query_type( $name, $query, $prefix ): string {
```
mindctrl, line 235: Thoughts on adding the param type declarations here? e.g. `string $name, array $query, string $prefix`

includes/Blocks/Block.php
```diff
@@ -237,72 +357,93 @@ private function resolve( $block ) {
...
-			$value          = null;
-			if ( empty( $rendered_block ) ) {
-				return $value;
-			}
-			switch ( $attribute_config['source'] ) {
+	private function resolve_block_attributes_recursive( $attributes, $html, $config ): array {
+		$result = [];
+		foreach ( $config as $key => $value ) {
+			// Get default value.
+			$default = isset( $value['default'] ) ? $value['default'] : null;
```
mindctrl, line 376:
```suggestion
			$default = $value['default'] ?? null;
```

includes/Type/Scalar/Scalar.php
```diff
@@ -32,4 +41,11 @@ public function init(): void {
 	public static function get_block_attributes_object_type_name(): string {
 		return 'BlockAttributesObject';
 	}
+
+	/**
+	 * Return type name of BlockAttributesArray.
+	 */
+	public static function get_block_attributes_array_type_name() {
```
mindctrl, line 48:
```suggestion
	public static function get_block_attributes_array_type_name(): string {
```

includes/Blocks/Block.php
theodesp: typing is hard :)

includes/Utilities/DomHelpers.php
```diff
@@ -23,12 +23,16 @@ final class DOMHelpers {
 	 *
 	 * @return string|null extracted attribute
 	 */
-	public static function parseAttribute( $html, $selector, $attribute, $default_value = null ): ?string {
+	public static function parseAttribute( $html, $selector = null, $attribute, $default_value = null ): ?string {
```
mindctrl, line 26: To match the phpdoc

```suggestion
	public static function parseAttribute( $html, $selector = '', $attribute, $default_value = null ): ?string {
```

includes/Utilities/DomHelpers.php
```diff
@@ -125,4 +129,28 @@ public static function getTextFromSelector( $html, $selector ) {
...
+	/**
+	 * Parses the html into DOMElement and searches the DOM tree for a given XPath expression or CSS selector.
+	 *
+	 * @param string      $html The HTML string to parse.
+	 * @param string|null $selector The selector to use.
+	 *
+	 * @return \DOMElement[]|\DOMElement
+	 */
+	public static function findNodes( $html, $selector = null ) {
+		$value = null;
```
mindctrl, line 142: My IDE is complaining about `$value` always being `null`. Could rename it to `$elem` to align with what's happening below.

includes/Blocks/Block.php
justlevine: I don't see a need to bump from 7.4 yet.

Like with union returns, if we _really need to use a mixed_, we can PHPDoc it. I'm on my phone and struggling to review the whole PR, but in this case im betting the actual type can be derived from the method itself.

includes/Blocks/Block.php
justlevine: @mindctrl @theodesp both PHPCompatibilityWP and phpstan should have caught this. I think there's an issue with the linter/ci 🤔

mindctrl (approved):
LGTM. I'm able to query data in tables as expected.

image

There are whitespace issues in the `CoreTableTest.php` file, with a mix of tabs and spaces.

mindctrl (approved):

.changeset/ten-dolphins-smash.md
```diff
@@ -0,0 +1,7 @@
+---
+"@wpengine/wp-graphql-content-blocks": minor
```
justlevine, line 2: The schema changes from `...Object` to `...Array` is breaking. 


```suggestion
"@wpengine/wp-graphql-content-blocks": major
```

mindctrl (approved):
Good catch on the major patch.

PR #181, opened 2024-01-25
Fix paragraph content default attribute

theodesp:
# Description

Adds a missing `default` value for the `CoreParagraph` and `CoreCode`  `content` fields.

# Testing

- Go to the GraphQL inspector and Search for the `CoreParagraphAttributes` type. It should be of type `String!` instead of `String`.
- Go to the GraphQL inspector and Search for the `CoreCodeAttributes` type. It should be of type `String!` instead of `String`.

# Screenshots
Screenshot 2024-01-25 at 15 18 53

### DIFF
--- a/.changeset/spotty-nails-complain.md
+++ b/.changeset/spotty-nails-complain.md
@@ -1,5 +1,5 @@
 ---
-"@wpengine/wp-graphql-content-blocks": patch
+"@wpengine/wp-graphql-content-blocks": major
 ---
 
-Adds missing default value for content attribute CoreParagraph block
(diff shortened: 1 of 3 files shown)


blakewilson (approved):
Works as expected! nice @theodesp. I did see the schema linter workflow failed, but I think we're still dealing with that as an issue right?

theodesp:
> Works as expected! nice @theodesp. I did see the schema linter workflow failed, but I think we're still dealing with that as an issue right?

The Schema check is correct. It checks against the latest tag released schema but we have breaking changes anyway.

PR #195, opened 2024-02-28
Fix updater slug mismatch

mindctrl:


blakewilson (approved):

blakewilson (changes requested):
There are some linting issues

blakewilson (approved):

PR #201, opened 2024-03-02
fix: rename `WPGraphQLHelpers` file to match class casing.

### LINKED ISSUE #168: Downloading the plugin as a composer package does not contain the vendor directory and autoloads
Hi

I recently upgraded my wp-graphql-content-blocks plugin from 1.1.1 to 2.0.0 and the first thing the graphql requests lead to is an exception saying that it can not find WPGraphQLHelpers. 
image

I checked in the downloaded plugin and saw no vendor/ folder. I checked the composer.json file for PSR-4 autoloads and was able to get it resolved via generating the autoloads for the plugin.

However, when I checked the zip present in the release https://github.com/wpengine/wp-graphql-content-blocks/releases/tag/v2.0.0, there is a vendor folder. So this issue happens only when loading this plugin via composer.

justlevine:
## What

This PR fixes the filename used for the `WPGraphQLHelpers` class to match the class casing.

This is *not* a breaking change, as this class is included via PSR-4 and not via `require_once`.

## Why

PSR-4 violation. Hopefully closes #168 🤞

## How
`includes/Utilities/WPGraphqlHelpers.php` => `includes/Utilities/WPGraphQLHelpers.php`.

No other changes have been made.


mindctrl (approved):
LGTM. Thank you @justlevine

PR #210, opened 2024-03-26
chore: stub WP_Post_Type and boostrap main plugin file during phpstan

justlevine:
## What

This PR

- Stubs `WP_Post_Type` so the custom GraphQL properties are available in PHPStan (and our IDEs)
- Includes `wp-graphql-content-blocks.php` in PHPStan's bootstrapper, to correctly identify the defined constants.

The PHPStan baseline has been updated accordingly.

## Why

Address false positives and uncover more _actual_ tech debt.


theodesp (approved):
LGTM!

PR #211, opened 2024-03-26
fix: check `post_content` before parsing it for blocks

justlevine:
## What

This PR fixes a PHP error where we were directly trying to access `post_content` on `get_post()` without verifying that an actual post is returned.

The logic has been changed to check if `$post->post_content` is empty, also saving us a call to `parse_blocks` if the post exists, but has no content.

> [!Warning]
> This PR requires #210 , which should be merged first (to avoid conflicts with the PHPStan baseline). The diff of specific changes is viewable here: a7aea08a1629ebbf8056e340f06a7d621babe42b

### DIFF
--- a/.changeset/quick-maps-exercise.md
+++ b/.changeset/quick-maps-exercise.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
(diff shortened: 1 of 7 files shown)


theodesp (approved):
LGTM

PR #212, opened 2024-03-26
fix: `get_allowed_block_types()` can return `bool`

justlevine:
## What

This PR changes the param type of `$supported_blocks_for_post_type_context` (used in `wpgraphql_content_blocks_should_apply_post_type_editor_blocks_interfaces` ) to type `array|bool`

> [!WARNING]
> ~This PR requires https://github.com/wpengine/wp-graphql-content-blocks/pull/210 , which should be merged first (to avoid conflicts with the PHPStan baseline). The diff of specific changes is viewable here: b5122b93f10535d4719086edc555a39f1625c21c~

## Why

`get_allow_block_types()` [can return a boolean](https://developer.wordpress.org/reference/functions/get_allowed_block_types/) when all block types are enabled/disabled

### DIFF
--- a/.changeset/neat-ties-prove.md
+++ b/.changeset/neat-ties-prove.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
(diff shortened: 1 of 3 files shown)


justlevine:
(rebased on `main`)

mindctrl (approved):

PR #217, opened 2024-03-27
chore: remove unnecessary `isset()` in `Anchor::get_block_interfaces()`

justlevine:
## What

This PR removes the unnecessary `isset( $block_spec )` check from the guard in `Anchor::get_block_interfaces()`

## Why

`$block_spec` is strictly-typed to `WP_Block_Type`, and therefore `isset()` will always be true.

## Additional notes

> [!IMPORTANT]
>


theodesp (approved):
LGTM!

PR #219, opened 2024-03-29
fix: make `get_block_attribute_fields()` `$prefix` param non-nullable

justlevine:
## What

This PR changes the `$prefix` parameter on `Blocks::get_block_attribute_fields()` from a nullable string (`string|null`) to a non-nullable `string`. The parameter remains optional and defaults to `''`.

>[!NOTE]
> This is _not_ considered a breaking change as `get_block_attribute_fields()` is a `private` class method, and [its only usage](https://github.com/justlevine/wp-graphql-content-blocks/blob/55fc5dc82a4bcaa4d4b8d78c286ca2306f66742d/includes/Blocks/Block.php#L90) has it passed a `string` value.

## Why

`Blocks::get_block_attribute_fields()` passes the `$prefix` param to `Blocks::get_attribute_type()` which requires the param to be a non-nullable string.

## Additional notes

> [!IMPORTANT]
>

### DIFF
--- a/.changeset/great-pans-poke.md
+++ b/.changeset/great-pans-poke.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+fix: Change Block:get_block_attribute_fields()`$prefix parameter be an optional`string`.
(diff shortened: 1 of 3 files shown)


theodesp (dismissed):
LGTM!

theodesp (approved):

PR #220, opened 2024-03-29
fix: handle arrays before casting when normalizing attribute values

justlevine:
## What
This PR fixes the logic inside `Block::normalize_attribute_values()` to:
- return `array` values before casting to the normalized `type`.
- cast strings to an array if the `$type` is an array, but the `$value` is not.

**No effort has been made** to determine the possible Scalar types that `$value` can accept, and instead is assumed to only be a `string|array` as currently defined in the doc-block.

## Why

You can't cast arrays to strings in PHP.

## Additional notes

> [!IMPORTANT]
>

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -292,10 +292,18 @@ class Block {
 	 * @param array|string $value The value
 	 * @param string       $type The type of the value
 	 * 
-	 * @return mixed
+	 * @return array|string|int|float|bool
 	 */
 	private function normalize_attribute_value( $value, $type ) {
+		// @todo use the `source` to normalize array/object values.
+		if ( is_array( $value ) ) {
+			return $value;
+		}
+
 		switch ( $type ) {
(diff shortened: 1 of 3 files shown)


includes/Blocks/Block.php
```diff
@@ -292,10 +292,18 @@ private function create_attributes_fields( $attributes, $prefix ): array {
...
 	 * @param string       $type The type of the value
 	 * 
-	 * @return mixed
+	 * @return array|string|int|float|bool
 	 */
 	private function normalize_attribute_value( $value, $type ) {
+		// @todo use the `source` to normalize array/object values.
+		if ( is_array( $value ) ) {
+			return $value;
+		}
```
mrclay, line 301: I think it would be clearer to move this into the `array` case block, and if $type is _not_ array, we shouldn't return an array, right?

theodesp (approved):

theodesp (approved):

PR #221, opened 2024-03-29
dev: inline and remove `Block::resolve()` and make `name` field nullable

justlevine:
## What
This PR:
- removes the `Block::resolve()` method, and instead inlines the `name` field resolver logic in `register_graphql_object_type()` $config.
- changes the `name` field resolver logic to return `null` if `$source['blockName']` is unset, instead of an empty string.

As the method is private, and the schema for the `name` field already being nullable, this is _not_ a breaking change.

## Why
Prompted by the lack of return type on `Block::resolve()`, I noted it's a private method (i.e. not overloadable by child classes), used only inside another private method `Block::register_type()`.  Removing the type both cleans up the tech debt, and makes it clearer to see what changes would be necessary to actually make `Block` and overloadable class.

## Additional notes

> [!IMPORTANT]
>

### DIFF
--- a/.changeset/serious-numbers-chew.md
+++ b/.changeset/serious-numbers-chew.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+dev: inline and remove `Block::resolve()` and make `name` field nullable.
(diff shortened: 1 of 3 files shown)


theodesp (approved):
LGTM!

PR #223, opened 2024-03-29
feat: Block object attributes may have custom types with subfields

### LINKED ISSUE #222: Can't set types of properties within an object attribute
Because WP doesn't define what an attribute with type `object` means, these attributes end up with the generic BlockAttributesObject type in this plugin, and can't be extended with `register_graphql_field('MyBlock', 'film', ...`. If you need decent types from blocks in GraphQL, then, you just can't use nested attributes. Nested attributes would improve the authoring experience of Save/Edit components, allowing us to build up structures from subcomponents.

One solution would be for the plugin to read object properties inside attributes like:

```
{
  ...
  "attributes": {
    "film": {
      "type": "object",
      "properties": {
        "id": {
          "type": "string"
        }
        ...
      }
    }
  }
}
```

A workaround would be to allow the field's GraphQL type to be overridden using `register_graphql_field` or some other function.

mrclay:
If a block attribute has a "properties" key, the attribute now has a custom type with typed subfields.

Fixes #222

Obviously a big concern here is that, to my knowledge, WP haven't documented how object attribute properties should be documented in block.json. "properties" was chosen by convention in JSONSchema.

### DIFF
--- a/.changeset/eight-steaks-yawn.md
+++ b/.changeset/eight-steaks-yawn.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": major
+---
+
+feat: Block object attributes may have custom types with subfields
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -164,7 +164,11 @@ class Block {
 					}
 					break;
 				case 'object':
-					$type = Scalar::get_block_attributes_object_type_name();
+                    if ( isset( $attribute['properties'] ) ) {
+                        $type = $this->get_query_type( $name, $attribute['properties'], $prefix );
+                    } else {
+                        $type = Scalar::get_block_attributes_object_type_name();
+                    }
 					break;
 			}
 		} elseif ( isset( $attribute['source'] ) ) {


justlevine:
> to my knowledge, WP haven't documented how object attribute properties should be documented in block.json. "properties" was chosen by convention in JSONSchema.

Some of the public docs are indeed a mess. ICYMI here's the docs for block.json [metadata](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-metadata/) and [attributes](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-attributes/). If there's something specific you're trying to verify I'm happy to help try and dig it out of the source code.

mrclay:
I think defining properties without a source is just undefined in WP and we should get their team to bless and document this approach before moving forward.

theodesp:
Hey @mrclay. Thank you for the PR. Before we review it we require a CLA agreement for you:

# Contributor License Agreement
All external contributors to WP Engine products must have a signed Contributor License Agreement (CLA) in place before the contribution may be accepted into any WP Engine codebase.

1. [Submit your name and email](https://wpeng.in/cla/)
2. 📝 Sign the CLA emailed to you
3. 📥 Receive copy of signed CLA

theodesp:
Hi @mrclay, thank you for the PR!

I’m not entirely sure if extending this to check for "properties" within "type": "object" is the best approach.

From my understanding, the block.json attributes aren't designed to follow a strict JSONSchema format or adhere to its rules. When "type": "object" is used, it primarily acts as a placeholder for any unstructured data type, which may or may not map to a valid query type—it could even represent jargon or binary data.

Would it be possible for you to keep this as a patch for now?

mrclay:
@theodesp The more general problem is #222 and I've tried to improve that issue's title/description. I'm not tied to this solution.

PR #224, opened 2024-03-30
fix: rename `DomHelpers.php` to `DOMHelpers.php` and improve type-safety of internal methods.

justlevine:
## What

This PR renames the `DomHelpers.php` file to `DOMHelpers.php` to match the class-name casing, per PSR-4.

Additionally, it fixes the internal and return types for the DomHelpers methods, and their usage.

## Why

`DiDom` methods have the ability to return either a `DiDom\Element` or a `DOMElement` instance depending on the supplied parameters. By explicitly typing those instances, we're able to unearth actual issues with the methods (like DOMHelpers::findNodes() returning `null` or being used as an array when it returns a single element.

## Additional notes

- The use of `DOMHelpers::findNodes()` in [Block::parse_single_source()](https://github.com/justlevine/wp-graphql-content-blocks/blob/e70bd1f014904eeb39d06134048f9024391b3141/includes/Blocks/Block.php#L457) is _not addressed_ in this PR. We probably want to be using `DOMHelpers::parseHTML()` (to correctly handle a possible array), but I didn't want to include a change I wasnt 100% clear on.

> [!IMPORTANT]
>
> To resolve without waiting for @justlevine , restore the base (i.e. `main` branch) `phpstan-baseline.neon` file, and then reapply the removal of the specific allow-listed error from the `phpstan-baseline.neon` diff in this PR.neon` diff in this PR.

### DIFF
--- a/.changeset/tame-baboons-juggle.md
+++ b/.changeset/tame-baboons-juggle.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+fix: rename `DomHelpers.php` to `DOMHelpers.php` and improve type-safety of internal methods.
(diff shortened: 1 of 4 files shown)


theodesp (approved):
LGTM!

PR #227, opened 2024-04-03
Only attempt to update schema on release tags

### LINKED ISSUE #197: Upload Schema Artifact workflow fails to upload schema
In #192 we attempted to fix the Schema Upload Artifact workflow by reverting [changes made previously](https://github.com/wpengine/wp-graphql-content-blocks/pull/155) which had resulted in the workflow not running. The attempted fix did not work. Source: https://github.com/wpengine/wp-graphql-content-blocks/actions/workflows/upload-schema-artifact.yml

The error is `GitHub Releases requires a tag`. It seems the context in which this runs does not have access to the tag data. cc @blakewilson for any additional context you may be able to add.

Every time we do a tagged release this workflow should run but it broke about 4 months ago and the schema linter job fails on every PR as a result. We've tried a change on this but it didn't fix it so we need to start over.

## Acceptance Criteria
- [ ] Ensure the schema linter can successfully fun on every PR and branch

## Notes
- https://github.com/wpengine/wp-graphql-content-blocks/actions/workflows/upload-schema-artifact.yml
- This will be difficult to test as it can only be tested on a release.
- Whomever works this should consider spinning up a separate repo for testing to avoid polluting this repo

ChrisWiegman:
This attempts to fix #197 by only running the schema generator on tag. The assumption is that the release workflow creates the tagged release itself so the release will be present and capable of receiving the upload.

Note that while this was tested in a fork of this repo, a lot of assumptions were made that prevented a full workflow from triggering. This can be further tested here by disabling the CircleCI runs and a willingness to delete extra releases from this repo.

### DIFF
--- a/.github/workflows/upload-schema-artifact.yml
+++ b/.github/workflows/upload-schema-artifact.yml
@@ -1,10 +1,9 @@
 name: Upload Schema Artifact
 
 on:
-  workflow_run:
-    workflows: ["Release"]
-    types:
-      - completed
+  push:
+    tags:
+      - "*"
 
 jobs:
   run:
@@ -22,8 +21,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
+
       - name: Setup WordPress
         uses: ./.github/actions/setup-wordpress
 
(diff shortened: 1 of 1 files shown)


theodesp (approved):
LGTM!

Issue #225, opened 2024-04-03
Missing fields in GraphQL query


MKlblangenois:
I recently upgrade WordPress core and plugins to the latest version, and now when I try to generate schemas, it gave me error about core blocks that missing some field (like `text` for `CoreButton` or `caption` for `CoreImage`)
![image](https://github.com/wpengine/wp-graphql-content-blocks/assets/42929225/03dbf291-f7ba-4f96-b1a7-8ede96bd873b)

When I inspect the schema.graphql from [the last release](https://github.com/wpengine/wp-graphql-content-blocks/releases/tag/v3.1.2), these fields exist but when I try to query them on GraphiQL IDE, they doesn't exist.
![SCR-20240403-irnt](https://github.com/wpengine/wp-graphql-content-blocks/assets/42929225/f4dc03b4-0dab-4757-822a-720c8237ea9d)

I've tried with default theme and other plugin disabled but the issue persist.

- Wordpress version: 6.5
- WPGraphQL: 1.22.1
- WPGraphQL Content Blocks: 3.1.2

MKlblangenois:
After several tests, it seems that the problem comes from the update to WordPress 6.5, I tested up to 6.4.3 without problem

theodesp:
@MKlblangenois it looks like WordPress does mess up their block fields with some breaking changes. At least GraphQL was able to detect any conflicts here before it's too late.

Hovever its quite tricky to rollout a new release of the blocks package since we will have to do a major upgrade release version to make this work. Clients could still be using old versions and if we update the fragments now it will break their code.

I could suggest to override the block fragments directly before you export them until we provide a new release.

Here is an example:
https://github.com/wpengine/faustjs/blob/canary/examples/next/block-support/wp-blocks/index.js
```
import { CoreBlocks } from '@faustwp/blocks';
import BlockB from './block-b/Component.js';
import { gql } from '@apollo/client';

CoreBlocks.CoreParagraph.fragments.entry = gql`

` // Update Fragment query here before exporting component.

export default {
  CoreParagraph: CoreBlocks.CoreParagraph,
  ...
````

theodesp:
I could also suggest to open  ticket in https://github.com/wpengine/faustjs so that we are aware of this change.

colis:
This is because the source for those blocks has changed to `rich-text`.

Adding `rich-text` as a case in the following switch should solve the problem

[...]

PR #229, opened 2024-04-04
Breaking Update: Handle `rich-text` field for WordPress >= v6.5

theodesp:
# Description
MAJOR: Update Schema to reflect latest WordPress 6.5 changes.

- **WHAT the breaking change is:** Added a new `rich-text` type.
- **WHY the change was made:** WordPress 6.5 replaced some of the attribute types from string to `rich-text` causing breaking changes to the existing block fields.
- **HOW a consumer should update their code:** If users need to use WordPress >= 6.5 they need to update this plugin to the latest version and update their graphql schemas.

# Related
https://github.com/wpengine/wp-graphql-content-blocks/issues/225

# Testing

-Update to WordPress 6.5
- Add a post with some paragraphs, code and heading blocks
- When querying their `content` attributes, there should be no issues or graphql errors.

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -136,9 +136,7 @@ class Block {
 
 		if ( isset( $attribute['type'] ) ) {
 			switch ( $attribute['type'] ) {
-				case 'rich-text': 
-					$type = 'String';
-					break;
+				case 'rich-text':
 				case 'string':
 					$type = 'String';
 					break;
@@ -300,7 +298,6 @@ class Block {
 	private function normalize_attribute_value( $value, $type ) {
 		switch ( $type ) {
 			case 'rich-text':
-				return (string) $value;
 			case 'string':
 				return (string) $value;
 			case 'number':
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -136,6 +136,8 @@ class Block {
 
 		if ( isset( $attribute['type'] ) ) {
 			switch ( $attribute['type'] ) {
+				case 'rich-text': 
+					$type = 'String';
 				case 'string':
 					$type = 'String';
 					break;
@@ -296,6 +298,8 @@ class Block {
 	 */
 	private function normalize_attribute_value( $value, $type ) {
 		switch ( $type ) {
+			case 'rich-text':
+				return (string) $value;
 			case 'string':
 				return (string) $value;
 			case 'number':
@@ -367,6 +371,7 @@ class Block {
 			$default = $value['default'] ?? null;
 			$source  = $value['source'] ?? null;
 			switch ( $source ) {
+				case 'rich-text':
 				case 'html':
 					if ( ! isset( $value['selector'] ) ) {
 						$result[ $key ] = $this->parse_single_source( $html, $source );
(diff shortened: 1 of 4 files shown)


includes/Blocks/Block.php
```diff
@@ -296,6 +299,8 @@ private function create_attributes_fields( $attributes, $prefix ): array {
 	 */
 	private function normalize_attribute_value( $value, $type ) {
 		switch ( $type ) {
+			case 'rich-text':
+				return (string) $value;
```
justlevine, line 303: This can pass-through like you do on 375

```suggestion
```

justlevine (commented):
Is there a pressing need to drop support for < 6.5 ? I'm not seeing anything that can't be done in a b/c way.

theodesp:
> Is there a pressing need to drop support for < 6.5 ? I'm not seeing anything that can't be done in a b/c way.

Yes it's tricky to do it nicely. It's annoying that we have to make a breaking change for some fields to work. We will have to document the supported versions so that users are aware.

justlevine:
There's a difference between breaking changes to the schema and dropping support for all previous WP versions, though. Not that either is really required from what I see, but if the Faust frontend is more paramount than this plugin, we can at least pipe the old data to the new schema with conditionals in the resolvers.

I know resources are tight rn, but I really believe it's worth it for WPE to handle this with care.

theodesp:
> There's a difference between breaking changes to the schema and dropping support for all previous WP versions, though. Not that either is really required from what I see, but if the Faust frontend is more paramount than this plugin, we can at least pipe the old data to the new schema with conditionals in the resolvers.
> 
> I know resources are tight rn, but I really believe it's worth it for WPE to handle this with care.

I think users with WordPress 6.4 should be able to use the current v3.0.0 package as it is. For users with WordPress 6.5 they will have to use the  updated version 4.0.0 of this plugin. Unless there is something I'm missing here.

ChrisWiegman:
@theodesp, I think David is right. We shouldn't break support for < 6.5 here, not less than 48 hours after it is out. What are our options on this?

theodesp:
This is the compatibility matrix as per own testing:


```
| **WordPress Version** | **wp-graphql-content-blocks** |
|-----------------------|-------------------------------|
| 6.4 (Before Fix)      | Compatible                    |
| 6.4 (After Fix)       | Compatible                    |
| 6.5 (Before Fix)      | Issues                        |
| 6.5 (After Fix)       | Compatible                    |
```

ChrisWiegman:
@justlevine, Given the further testing Theo did, is there something I'm missing here as it relates to this plugin? Doesn't look like we'll have to worry about backwards compatibility?

justlevine:
Upon further testing @theodesp is right. I thought I was getting errors from the removal of `content`, but it was actually because I forgot to rebuild my composer deps after #224 🤦

(Someone should probably open up a separate issue about improving the coverage of unit tests...)

ChrisWiegman:
Thanks, @justlevine , and I couldn't agree more on the test coverage. It's actually something Theo and I had chatted about last week.

ChrisWiegman (approved):

PR #231, opened 2024-04-04
Adds dev env artifacts to .gitignore as appropriate

ChrisWiegman:
As I use a docker-based dev environment similar to Lando, this ensures its files are properly ignored to avoid issues later.


theodesp (approved):
LGTM!

PR #232, opened 2024-04-08
Add blocks unit tests

theodesp:
# Description

Adds Smoke Unit tests for the following blocks:

* CoreCode
* CoreHeading
* CoreList
* CoreButton
* CoreQuote

### DIFF
--- a/.changeset/famous-hats-deliver.md
+++ b/.changeset/famous-hats-deliver.md
@@ -1,0 +1,9 @@
+---
+"@wpengine/wp-graphql-content-blocks": major
+---
+
+MAJOR: Update Schema to reflect latest WordPress 6.5 changes.
+
+- WHAT the breaking change is: Added new `rich-text` type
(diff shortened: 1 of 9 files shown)


justlevine:
Not sure where to share this, and it's probably beyond the scope of this particular, but I recommend using the official [block theme test data](https://github.com/WordPress/theme-test-data/blob/master/64-block-test-data.xml) as a reference.

theodesp:
> Not sure where to share this, and it's probably beyond the scope of this particular, but I recommend using the official [block theme test data](https://github.com/WordPress/theme-test-data/blob/master/64-block-test-data.xml) as a reference.

Interesting. Wouldn't that require parsing the xml data and inserting certain author and post ids though? I think we could add an improvement ticket by adding a unit test step to load the block theme test data in the database once and then run queries based on those.

theodesp:
Closing as already implemented

Issue #239, opened 2024-05-09
CoreTable column alignment returns null


brysonchiu:
The core table block is returning null values for cell alignment even when it is set. The cell alignment is instead set by the whole table alignment. This seems like a mistake.

**Steps to reproduce**

1. Add a core table into a page and adjust the alignment on a column (not the alignment of the table).
2. Add this into your page query:
3. 
```javascript
    editorBlocks {
      ... on CoreTable {
        attributes {
          body {
            cells {
              align
            }
          }
          foot {
            cells {
              align
            }
          }
          head {
            cells {
              align
            }
          }
        }
      }
    }
```

theodesp:
Hey @phillubliner I've reproduced this issue and I will provide a fix soon.

theodesp:
@phillubliner @phillubliner

https://github.com/wpengine/wp-graphql-content-blocks/releases/tag/v4.0.1

Should contain a fix for this issue. I'm closing this ticket but feel free to comment back if you still have any problems.

MajesticPagan:
Hello @theodesp, I can confirm the same: `content` still returns `null`. Is there any news on this? Thanks!

lekkacat:
I think the **content** problem is here:
https://github.com/wpengine/wp-graphql-content-blocks/blob/7c3a6d7e2b0825f9e3cdd9e6411152d4957bb114/includes/Blocks/Block.php#L139


The content of the table cells is in the format `<TD>TheContent</TD>`and must be parsed as html but the source is rich-text.  If i change to 
`case 'rich-text':
$result[$key] = $this->parse_single_source($html, 'html');
break;`

The cells are populated but I'am not deep enough in the topic to understand if this will course strange functions in other attributes.

brysonchiu:
@theodesp

I have just updated to the latest WP, 6.6.1, and the latest wp-graphql-content-blocks, 4.1.0, and the cell alignment is still returning null. Im updating from WP 6.4.5 and wp-graphql-content-blocks 3.1.2.

Additionally, like others have stated, the content is too returning null. This is a new issue for me since updating.

Should I open a new issue or continue with this thread?

PR #243, opened 2024-06-13
Fix parseAttribute Function to Handle Multiple Nodes

theodesp:
# Description 
This PR addresses a bug in the`parseAttribute` function where it only checked the first node for the specified attribute when the selector was empty. The function has been updated to iterate through all nodes that match the selector and return the attribute value if found. Additionally, unit tests have been updated and expanded to cover these changes.

# Fixes
https://github.com/wpengine/wp-graphql-content-blocks/issues/239


ChrisWiegman (approved):
LGTM

PR #244, opened 2024-06-15
ci: add PHPStan workflow

justlevine:
## What

This PR runs PHPStan as part of the CI workflow on pull requests.

## Why

Prevent regressions in static analysis.


justlevine:
Does a change to the CI and not the actual code require a changeset?

theodesp:
Thanks @justlevine. It should be fine without changesets.

theodesp (approved):

PR #245, opened 2024-06-15
fix: undefined array keys in `Block::resolve_block_attributes_recursive()`

### LINKED ISSUE #238: ApolloError: Internal server error when querying "dropCap" attribute on core/paragraph block
When I query core/paragraph block with full list of attributes, I am getting the following FE error: 

```javascript
error - GraphQL errors: [
  {
    debugMessage: 'Undefined array key "dropCap"',
    message: 'Internal server error',
    extensions: { category: 'internal' },
    locations: [ [Object] ],
    path: [ 'page', 'editorBlocks', 0, 'attributes', 'dropCap' ]
  }
]
```

and GraphQL error: 

```javascript
"errors": [
  {
    "debugMessage": "Undefined array key \"dropCap\"",
    "message": "Internal server error",
    "extensions": {
      "category": "internal"
    },
    "locations": [
      {
        "line": 20,
        "column": 7
      }
    ],
    "path": [
      "page",
      "editorBlocks",
      0,
      "attributes",
      "dropCap"
    ]
  }
]
```

In the Wp Admin GraphQL IDE, `dropCaps` attributed is in the list of available attributes. 

 
Seems like the issue is only with the attribute `dropCap`. When it is not in the query, all works as expected. This behaviour is on: 
- WordPress 6.5.2
- WPGraphQL Content Blocks 4.0.0

**Additional relevant plugins and packages used on FE:**
```javascript 
"@apollo/client": "^3.9.11",
"@faustwp/blocks": "^4.0.0",
"@faustwp/cli": "^3.0.1",
"@faustwp/core": "^3.0.1", 
"@wordpress/base-styles": "^4.47.0",
"@wordpress/block-library": "^8.33.0",
"@wordpress/style-engine": "^1.39.0",
"@wordpress/scripts": "^27.7.0",
"@faustwp/block-editor-utils": "^0.2.0",
```

**Additional relevant plugins and packages used on BE:**
```javascript
"wpackagist-plugin/faustwp":"1.2.3",
"wpackagist-plugin/wp-graphql":"1.23.0",
"wpackagist-plugin/wpgraphql-acf":"2.2.0",
"wpackagist-plugin/wpgraphql-smart-cache":"1.3.1",
"wpengine/advanced-custom-fields-pro": "^6.2.9",
"wpengine/wp-graphql-content-blocks": "^4.0.0",
```

Any idea what is going on? Could https://github.com/wpengine/wp-graphql-content-blocks/pull/182 be relevant to this issue?

justlevine:
## What

This PR refactors `Block::resolve_block_attributes_recursive()` in the following ways:

1. Add types to the method signature _when the types are certain_.
2. Breaks out source parsing into individual methods, for reduced code complexity.
3. Checks for array keys before they are used on the array:
- On required `$value` keys that are necessary to parse the source.
- Before post-processing the `$return[ $key ]` values retrieved from the source.

## Why
Prevents `Undefined array key` PHP warnings, as well as improves handling of malformed blocks.

Fixes #238

### DIFF
--- a/includes/Blocks/Block.php
+++ b/includes/Blocks/Block.php
@@ -366,73 +366,60 @@ class Block {
 	/**
 	 * Resolved the value of the block attributes based on the specified config
 	 * 
-	 * @param array  $attributes The block current attributes value
-	 * @param string $html The block rendered html
-	 * @param array  $config The block current attribute configuration
+	 * @param array<string,mixed> $attributes The block current attributes value.
+	 * @param string              $html The block rendered html.
+	 * @param array<string,mixed> $config The block current attribute configuration, keyed to the attribute name.
 	 */
-	private function resolve_block_attributes_recursive( $attributes, $html, $config ): array {
+	private function resolve_block_attributes_recursive( $attributes, string $html, array $config ): array {
 		$result = [];
+
 		foreach ( $config as $key => $value ) {
 			// Get default value.
 			$default = $value['default'] ?? null;
 			$source  = $value['source'] ?? null;
+
 			switch ( $source ) {
 				case 'rich-text':
 				case 'html':
+					// If there is no selector, we are dealing with single source.
 					if ( ! isset( $value['selector'] ) ) {
 						$result[ $key ] = $this->parse_single_source( $html, $source );
-					} else {
-						$result[ $key ] = DOMHelpers::parseHTML( $html, $value['selector'] );
-	
-						if ( isset( $value['multiline'] ) && ! empty( $result[ $key ] ) ) {
-							$result[ $key ] = DOMHelpers::getElementsFromHTML( $result[ $key ], $value['multiline'] );
-						}
+						break;
 					}
+
+					$result[ $key ] = $this->parse_html_source( $html, $value );
 					break;
 				case 'attribute':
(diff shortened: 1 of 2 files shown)


includes/Blocks/Block.php
```diff
@@ -366,73 +366,60 @@ private function register_type(): void {
...
+						$result[ $key ] = true;
 						break;
-					}
-					$result[ $key ] = true;
-					break;
+				}
 			}
-			
+
+			// Fallback to the attributes or default value if the result is empty.
```
justlevine, line 422: Now that this method is cleaned up, we can more easily see that the `$result[ $key ]` is getting replaced with the fallback or default value even when it probably shouldn't: 
E.g. an `int` value of 0, or a boolean value of `false`.

I didn't want to change this behavior because while I doubt it's intentional, I'm not 100% sure and didn't want to introduce a breaking change.

theodesp (approved):

includes/Blocks/Block.php
justlevine: @theodesp @ChrisWiegman when one of you has a minute can you confirm whether this is a bug or intended behavior?

PR #256, opened 2024-07-23
chore: update Composer dev-dependencies and fix resulting issues.

justlevine:
## What

This PR updates the Composer dev-dependencies to their latest versions, and fixes the newly exposed PHPCS/PHPStan smells.

## Why

Got a bunch of incoming PRs lined up, and want to use accurate tooling.

## How

The only production-code change in this PR is fixing the recursive use of `Block::get_attribute_type()` from static to instance, since it's registered as a private instance method on the class.


justlevine:
For your review @jasonbahl

theodesp (approved):

PR #258, opened 2024-08-02
ci: replace `docker-compose` commands with `docker compose`

justlevine:
## What

This PR replaces all instances of the outdated `docker-compose` command with the current syntax: `docker compose`

## Why

Havn't seen an article yet, but seems like GH runners dropped compatibility with the old command: https://github.com/wpengine/wp-graphql-content-blocks/actions/runs/10094219877/job/28273101015?pr=257


theodesp:
@justlevine thanks I will trigger a run to see if it works.

theodesp (approved):

PR #267, opened 2024-09-01
chore: update composer dev-deps

justlevine:
## What

Updates Composer dev-dependencies to their latest (semver-compatible) versions.

## Why

New version of PHPStan


justlevine:
Not sure why this would conflict when it's based on `main` 🤔

This pr can be discarded in favor of someone else running and committing `composer update -W`

theodesp (approved):

jasonbahl (approved):

PR #279, opened 2024-09-12
dev: refactor ContentBlocksResolver and deprecate `TraverseHelpers`

justlevine:
## What

This PR refactors `ContentBlocksResolver` for code complexity/maintainability (SOLID) and to better resembles WordPress's internal block parsing/rendering.

More specifically, `::resolve_content_blocks()` no longer retains any top-level logic, and instead blocks are parsed recursively (`::parse_blocks() -> `::handle_do_blocks()` -> `::handle_do_block()` )

As a result, `TraverseHelpers` has been deprecated entirely.

## Why

This lays the groundwork for fixing missing values due to incorrect hydration, moving field resolution from resolvers to a model, fixing layout css classes, and a few performance improvement strategies I'm investigating.

> [!NOTE]
>
> ~There is _no_ pressing need to merge this until either more tests are backfilled or future PRs that actually address code issues are pushed for review.~
> Followed up by #284


includes/Utilities/TraverseHelpers.php
```diff
@@ -7,21 +7,37 @@
...
 	 * @param int      $depth The current depth of traversal.
 	 * @param int      $max_depth The maximum depth to traverse.
 	 */
 	public static function traverse_blocks( &$blocks, $callback, $depth = 0, $max_depth = PHP_INT_MAX ): void {
+		_deprecated_function(
+			__METHOD__,
+			'@todo',
+			sprintf(
+			// translators: %s: class name
+				esc_html__( 'Manipulate blocks directly inside %s::handle_do_block', 'wp-graphql-content-blocks' ),
```
jasonbahl, line 36: missing sprintf?

justlevine: In addition to the one on L34?

theodesp (approved):

includes/Utilities/TraverseHelpers.php
jasonbahl: nah we're good. i think it was a presentation thing in the Github diff. I see it now.

PR #292, opened 2024-09-23
tests: backfill tests for `ContentBlockResolver`

Ta5r:
## What
This PR backfills additional test cases for the `ContentBlocksResolver` class in `ContentBlocksResolverTest`.

## How

The following test cases are included:   

   - **Pre-resolve Filter Test:** `test_pre_resolved_blocks_filter_returns_non_null` ensures that the `wpgraphql_content_blocks_pre_resolve_blocks` filter correctly modifies the block content before the blocks are resolved.
   - **Empty Content Test:** `test_returns_empty_array_for_empty_content` checks that an empty array is returned when the post content is empty, ensuring proper behavior for posts with no blocks.
   - **Post-resolve Filter Test:** `test_filters_wpgraphql_content_blocks_resolve_blocks` ensures that the `wpgraphql_content_blocks_resolve_blocks` filter correctly modifies block content after the blocks are resolved.
   - **Inner Block Test:** `test_inner_blocks` tests resolution for both flat and nested blocks, including the correct `parentClientIds` are set, and that there are no discrepancies between flat/nested resolution.

### DIFF
--- a/.changeset/weak-days-suffer.md
+++ b/.changeset/weak-days-suffer.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+tests : backfill tests for ContentBlockResolver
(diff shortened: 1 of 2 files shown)


Ta5r:
c.c @justlevine

jasonbahl (approved):

PR #293, opened 2024-09-23
fix: Correctly parse nested attribute and tag sources.

### LINKED ISSUE #282: Bug: CoreTable doesn't return most data for fields
Most of the data for `CoreTable` returns `null`.  The only meaningful data is returned in the `renderedHtml` field, which is not ideal.

Here's the query I used:

```
{
  pages {
    nodes {
      title
      editorBlocks(flat: false) {
        name
        clientId
        parentClientId
        ... on CoreTable {
          anchor
          apiVersion
          attributes {
            textColor
            style
            metadata
            lock
            head {
              cells {
                align
                colspan
                content
                rowspan
                tag
                scope
              }
            }
            hasFixedLayout
            gradient
            foot {
              cells {
                align
                colspan
                content
                rowspan
                scope
                tag
              }
            }
            fontSize
            fontFamily
            className
            caption
            borderColor
            body {
              cells {
                align
                colspan
                content
                rowspan
                scope
                tag
              }
            }
            backgroundColor
            anchor
            align
          }
          renderedHtml
        }
      }
    }
  }
}
```

justlevine:
## What

This PR fixes an issue with nested attributes (such as those used by CoreTable) not being resolved correctly.

More specifically, it:
- Ensures `::parse_attribute()` is called correctly regardless of if there is a `selector` present.
- Adds support for parsing `tag` sources
- Removes the unnecessary `::parse_single_source()

Additionally, tests have been backfilled specific CoreTable attributes.

## Why

Fixes #282 

### Covered `CoreTableAttributes`

- align
- backgroundColor
- borderColor
- caption
- className
- fontFamily
- fontSize
- gradient
- hasFixedLayout
- lock
- style
- textColor
- head
- foot
- body

### Not covered
- `CoreTableAttributes.metadata`

## Additional notes:

The new `DOMHelpers::get_first_node_tag_name()` is marked as internal, because the entire class needs to be reworked (and ideally replaced with WordPress's HTML parsing api ).


theodesp:
Ta!

theodesp (approved):

PR #295, opened 2024-09-23
tests: Add tests for `CoreHeading` block

ashutoshgautams:
## What
This PR backfills tests for the `CoreHeading` block and its attributes.

### Tested attributes:
- align
- anchor
- backgroundColor
- className
- content
- cssClassName
- fontSize
- gradient
- level
- style
- textAlign
- textColor

### Untested fields:
- `CoreHeadingAttributes.metadata` - @todo
- `CoreHeadingAttributes.lock` - Not explicitly tested, always returns null in current tests
- `CoreHeadingAttributes.placeholder` - Not explicitly tested, always returns null in current tests

### Exposed issues:
- `CoreHeading.cssClassNames`: returns `null` even when the block has CSS classes. This is commented out in the test file.
- `CoreHeadingAttributes.level` returns a `Float` instead of an `Integer`. This is noted in a comment in the test file.

These issues are noted in comments within the test file and should be addressed in future updates.


ashutoshgautams:
cc: @justlevine

PR #316, opened 2024-11-28
fix: Ensure correct `EditorBlock.type` field resolution

justlevine:
## What

This PR fixes a regression in #285 where the `type` field was exposed with no resolve method, cause the field to always return null.

## Why

In the upstream branch I'm hydrating blocks with a model, but we need an explicit resolve function until that's merged.


theodesp (approved):

theodesp:
Ta!

PR #323, opened 2024-12-17
feat: add support for resolving pattern blocks

justlevine:
## What

This PR adds support to resolving the `innerBlocks` for synced Block Patterns, introduced in WordPress 6.6

## Why
While synced patterns are a highlight of FSE, they are still commonly inserted manually into post/page/cpt content, especially in hybrid themes.

Since this is a backwards-compatible improvement, I believe it makes sense to introduce this now instead of after any incoming breaking changes.

## How

See diff


colinmurphy (approved):
Thanks @justlevine 😁

PR #324, opened 2024-12-17
feat: add support for resolving Template Part blocks

justlevine:
## What

This PR adds support for resolving `innerBlocks` of Template Parts, introduced in WP 5.8

## Why

While less common outside of FSE, template parts are still commonly used by hybrid themes and DIY devs to inject reusability into individual posts/pages. Since this is a backwards-compatible change, it makes sense to handle this before any upcoming breaking releases.


theodesp:
Awesome!

justlevine:
Rebased to fix the merge conflict.

theodesp (approved):

PR #331, opened 2025-01-14
Feature: Added CoreListItem Block and updated CoreList block

colinmurphy:
# Overview

This is a solution for the issue reported #253  whereby a user might want to recursively query child blocks.

This PR has 2 main features

1. Adds a CoreListItem Block
2. Updates the CoreList Block with new fields so that you can recursively get children for the CoreListItem

# Example

You have a list with 3 levels  e.g.

Screenshot 2025-01-14 at 12 53 00

The HTML for the block is

```
<!-- wp:list {"ordered":true} -->
<ol class="wp-block-list"><!-- wp:list-item -->
<li>List item 1</li>
<!-- /wp:list-item -->

<!-- wp:list-item -->
<li>List item 2<!-- wp:list {"className":"is-style-checkmark-list"} -->
<ul class="wp-block-list is-style-checkmark-list"><!-- wp:list-item -->
<li>Child list item 1<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>Third level list item</li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item -->

<!-- wp:list-item -->
<li>Child list item 2</li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item -->

<!-- wp:list-item -->
<li>List item 3</li>
<!-- /wp:list-item -->

<!-- wp:list-item -->
<li>List item 4</li>
<!-- /wp:list-item --></ol>
<!-- /wp:list -->
```

The following query will do the following

1. Check if the list is ordered (boolean)
2. Get an array of items (CoreListItem blocks) and get children and values for those blocks and child blocks

```gql
query postQuery($id: ID!) {
  post(id: $id, idType: DATABASE_ID, asPreview: false) {
    title
    editorBlocks(flat: false) {
      name
      ... on CoreList {
        ordered
        items {
          value
          children {
            value
            children {
              value
            }
          }
        }
      }
    }
  }
}
```

The response from WP GraphQL is

```json
{
  "data": {
    "post": {
      "title": "Hello world!",
      "editorBlocks": [
        {
          "name": "core/list",
          "ordered": true,
          "items": [
            {
              "value": "<li>List item 1</li>",
              "children": []
            },
            {
              "value": "<li>List item 2</li>",
              "children": [
                {
                  "value": "<li>Child list item 1</li>",
                  "children": [
                    {
                      "value": "<li>Third level list item</li>"
                    }
[...]

PR #334, opened 2025-01-16
feat: Add support for CoreNavigation blocks

theodesp:
# Summary
This PR adds support for resolving and returning navigation items within the CoreNavigation block for WPGraphQL Content Blocks. The implementation parses the referenced navigation post and returns a list of block types (EditorBlock) representing the inner blocks of the navigation.

### Note 
The implementation does not consider the list of allowed blocks so we do not do any validation checks as we rely on Gutenberg logic.

## Usage Example
With this update, the GraphQL query below is now supported:

```graphql
{
  posts {
    nodes {
      editorBlocks {
        ... on CoreNavigation {
          type
          name
          innerBlocks {
            type
            name
          }
          attributes {
            ref
          }
        }
      }
    }
  }
}
```
The `navigationItems` field will return a list of parsed blocks from the referenced navigation post, which includes nested block types like `CoreNavigationLink`.

Example Response
```json
{
  "data": {
    "posts": {
      "nodes": [
        {
          "editorBlocks": [
            {
              "type": "CoreNavigation",
              "name": "core/navigation",
              "innerBlocks": [
                {
                  "type": "CorePageList",
                  "name": "core/page-list"
                },
                {
                  "type": "CoreNavigationLink",
                  "name": "core/navigation-link"
                }
              ],
              "attributes": {
                "ref": 31
              }
            },
            {},
            {},
            {}
          ]
        },
        {
          "editorBlocks": [
            {}
          ]
        }
      ]
    }
  },
}
```


justlevine (commented):
My personal opinion here is the same as on https://github.com/wpengine/wp-graphql-content-blocks/pull/331#pullrequestreview-2553377033 - and will be for any time a PR
- registers a custom field to a Block (maintainability, forward-compat).
- calls `render_block()` (performance,  bugs from rerendering)

Doubly true here, when (like with CoreList), `CoreNavigationLink` is **already available** using the proper pattern of `innerBlocks`.

image

colinmurphy (approved):

includes/Blocks/CoreNavigation.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	/**
+	 * Filter out empty blocks from the parsed blocks.
+	 *
+	 * @param array $parsed_blocks An array of parsed block data.
+	 * @return array An array of filtered blocks.
+	 */
+	private static function filter_out_empty_blocks( array $parsed_blocks ): array {
+		$filtered = array_filter(
```
whoami-pwd, line 68: I am not sure if we need the intermediate variable `$filtered` here.

theodesp:
> My personal opinion here is the same as on [#331 (review)](https://github.com/wpengine/wp-graphql-content-blocks/pull/331#pullrequestreview-2553377033) - and will be any time a PR:
> 
> * registers a custom field to a Block (maintainability, forward-compat).
> * calls `render_block()` (performance,  bugs from rerendering)
> 
> Doubly true here, when (like with CoreList), `CoreNavigationLink` is **already available** using the proper pattern of `innerBlocks` / `flatListToHierarchical()`.
> 
> image

TBH innerBlocks does not render anything for me:

```
{
  "data": {
    "posts": {
      "nodes": [
        {
          "editorBlocks": [
            {
              "innerBlocks": [],
              ...
}

{
  "data": {
    "posts": {
      "nodes": [
        {
          "editorBlocks": [
            {
              "innerBlocks": [],
 ```
 
 IF you look at the actual resolved block data its just this:
 
 ```
 {
  "blockName": "core/navigation",
  "attrs": {
    "ref": 31
  },
  "innerBlocks": [],
  "innerHTML": "",
  "innerContent": [],
  "clientId": "67892bb3d0bb0"
}
```

So if we could expose the list of innerBlocks by just returning the parsed list of blocks from the CoreNavigation. Is there a better way you think?

includes/Blocks/CoreNavigation.php
```diff
@@ -0,0 +1,78 @@
...
+			$parsed_blocks,
+			static function ( $block ) {
```
whoami-pwd, line 70: Maybe we can use `fn` instead, like

```suggestion
			static fn ( array $block ): bool => ! empty( $block['blockName'] )
```

justlevine:
> So if we could expose the list of innerBlocks by just returning the parsed list of blocks from the CoreNavigation. Is there a better way you think?

If it's not working on `post.editorBlocks`, then it's a global state issue (probably in in https://github.com/rtCamp/wp-graphql-content-blocks/pull/31 that I havn't backported yet 😬).

My recommendation of how to handle it in the interim would be to populate it `ContentBlocksResolver`, similar to how we currently handle reusable blocks: https://github.com/wpengine/wp-graphql-content-blocks/blob/b133a1b36d9e5bfb407e46931c0c279b7a62dc43/includes/Data/ContentBlocksResolver.php#L252C1-L270C3

whoami-pwd (changes requested):
Just a couple of minor code related issues here that we can improve.

theodesp:
> https://github.com/wpengine/wp-graphql-content-blocks/blob/b133a1b36d9e5bfb407e46931c0c279b7a62dc43/includes/Data/ContentBlocksResolver.php#L252C1-L270C3

Thanks. I will update the code then.

includes/Blocks/CoreNavigation.php
```diff
@@ -0,0 +1,78 @@
...
+		$attributes = $block['attrs'] ?? [];
+		$ref        = $attributes['ref'] ?? null;
+
+		if ( ! $ref ) {
+			return null;
+		}
+
+		$navigation_post = get_post( $ref );
+		if ( ! $navigation_post || 'publish' !== $navigation_post->post_status ) {
+			return [];
```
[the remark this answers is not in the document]
theodesp: An empty list is correct. I will fix the  null return.

includes/Blocks/CoreNavigation.php
```diff
@@ -0,0 +1,78 @@
...
+	public function resolve_navigation_items( $block ) {
+		$attributes = $block['attrs'] ?? [];
+		$ref        = $attributes['ref'] ?? null;
+
+		if ( ! $ref ) {
+			return null;
+		}
+
+		$navigation_post = get_post( $ref );
+		if ( ! $navigation_post || 'publish' !== $navigation_post->post_status ) {
```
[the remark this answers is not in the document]
theodesp: Actually, its better if we do what gutenberg does in that case:

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/navigation/index.php#L230-L237

I will update the code

includes/Blocks/CoreNavigation.php
theodesp: A neat trick: By assigning the result of this operation to a variable, you can set breakpoints or inspect the state of $filtered during debugging.

I could remove it to be more concise of course.

colinmurphy (approved):

includes/Data/ContentBlocksResolver.php
```diff
@@ -242,6 +243,33 @@ private static function populate_post_content_inner_blocks( array $block ): arra
...
 
+	/**
+	 * Populates the innerBlocks of this block with navigation item blocks from the referenced navigation post.
+	 *
+	 * @param array<string,mixed> $block The block to populate.
+	 *
+	 * @return array<string,mixed> The populated block.
+	 */
+	private static function populate_navigation_blocks( array $block ): array {
+		if ( 'core/navigation' !== $block['blockName'] || ! isset( $block['attrs']['ref'] ) ) {
```
whoami-pwd, line 254: If we are going to rely on a strict types it would be beneficial for us to pay attention when using the undescribed array.
Maybe something like this can prevent from potential notice.

```suggestion
		if ( ($block['blockName'] ?? '') !== 'core/navigation' || ! isset( $block['attrs']['ref'] ) ) {
```

includes/Data/ContentBlocksResolver.php
```diff
@@ -242,6 +243,33 @@ private static function populate_post_content_inner_blocks( array $block ): arra
...
+			return $block;
+		}
+
+		$navigation_post = get_post( $block['attrs']['ref'] );
```
whoami-pwd, line 258: Should we check vs type or just typecast the `$block['attrs']['ref']` ?
int|WP_Post

includes/Data/ContentBlocksResolver.php
```diff
@@ -242,6 +243,33 @@ private static function populate_post_content_inner_blocks( array $block ): arra
...
+
+		if ( ! $navigation_post || 'publish' !== $navigation_post->post_status ) {
```
whoami-pwd, line 260: Can we check if the `$navigation_post instanceof WP_Post` here instead of `! $navigation`?

whoami-pwd (commented):
Nice work 👍🏻 
Just a couple of minor suggestions.

whoami-pwd (approved):

includes/Data/ContentBlocksResolver.php
theodesp: I don't think this is necessary since this is how Gutenberg checks it as well

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/navigation/index.php#L230-L241

includes/Data/ContentBlocksResolver.php
theodesp: I don't think this is necessary since this is how Gutenberg checks it as well

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/navigation/index.php#L230-L241

justlevine: I would cast it to `(int)` just so nobody needs to monitor if wp decides to change the `ref` to a numeric string.

includes/Data/ContentBlocksResolver.php
justlevine: @whoami-pwd / @theodesp I strongly disagree with this change, if `blockName` is not set at this point in the lifecycle we have bigger problem that we shouldn't be hiding behind null coercion.

https://github.com/wpengine/wp-graphql-content-blocks/blob/3bbe3e4a033ae664d1a95701bb065a3284acbcf8/includes/Data/ContentBlocksResolver.php#L138C1-L148C33

whoami-pwd: Actually that is a good point 👍🏻 
Not entirely agree on the default notice here, I would say definitely something to think about for future.

includes/Data/ContentBlocksResolver.php
whoami-pwd: With all due respect to the Gutenberg code sample here, we are using the `$navigation_post` as an instance of a WP_Post on next lines. If we decide to tighten things up with phpstan in future that might be an issue as.

colinmurphy (approved):

Issue #180, opened 2024-01-24
CorePostTerms doesn't expose any way to ask for the terms


jasonbahl:
When querying Blocks, if a block is of the `CorePostTerms` block type, there's no way on the block to fetch the terms.

In the Block Editor I can see the terms: 

![CleanShot 2024-01-24 at 15 55 40](https://github.com/wpengine/wp-graphql-content-blocks/assets/1260765/f8d0c0e8-0d97-400a-b0a7-8cefc3332565)

However when querying for blocks, I cannot find any field in the Schema that would allow me to fetch these terms and make use of them. 

I would like to be able to query the terms as a connection. 

Something to the tune of: 

```graphql
query WysiwygField($uri: String! = "my-uri") {
  nodeByUri(uri: $uri) {
    id
    uri
    ...on NodeWithEditorBlocks {
      editorBlocks {
        __typename
        ...on CorePostTerms {
          ## There's no way to query the terms here 🤔
          terms { ## Would be great to be able to query like this
            nodes {
              __typename
               id
               name
            }
          }
        }
      }
    }
  }
}
```

theodesp:
Hey @jasonbahl  is CorePostTerms a Block type or we have to create an artificial type there?

justlevine:
@jasonbahl afaik no query blocks are currently supported by this plugin. and @theodesp I'd be wary of creating an Artificial type to support it without planning for forward compatibility with the [Block Bindings API](https://github.com/WordPress/gutenberg/issues/54536) that we'll hopefully start seeing the impacts of in WP 6.5, and when completed should provide the missing pieces we need for FSE.

theodesp:
@jasonbahl is there a way to use the Selector API to define an array attribute for this?

With [this PR](https://github.com/wpengine/wp-graphql-content-blocks/pull/176) we support array types

So it works similar with CoreTable block

jasonbahl:
@theodesp not sure, haven't looked into it yet. 

Just was trying to use the block and determined it was essentially useless for me at the moment so wanted to open an issue to discuss. I imagine anyone else using blocks in a headless environment might find this block difficult to use as-is. 🤷🏻‍♂️ 

For the moment, I'm using an ACF Block and an ACF Field on the block to solve my current need of being able to query a list of terms from a block.

theodesp:
@jasonbahl do you think we can include an additional field to resolve the Post Terms similar to the CoreImage block?

[...]

PR #339, opened 2025-02-05
feat: Adds terms querying functionality for the CorePostTerms Block.

whoami-pwd:
### Description

Related issue: #180 

- Adds terms querying functionality for the CorePostTerms Block.
New `terms` ~~field~~ **connection** registered. ~~Resolves as a list_of TermNode~~
New `taxonomy` connection registered.

- Additionally adds support for the `prefix` and `suffix` items with correspondent newly registered string fields.

```graphql
query TestPostTerms($uri: String! = "test-terms") {
  nodeByUri(uri: $uri) {
    id
    uri
    ... on NodeWithPostEditorBlocks {
      editorBlocks {
        __typename
        ... on CorePostTerms {
          prefix
          suffix
          taxonomy {
            __typename
            node {
              __typename
              id
              name
            }
          }
          terms {
            __typename
            nodes {
              __typename
              id
              name
            }
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/includes/Blocks/CorePostTerms.php
+++ b/includes/Blocks/CorePostTerms.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Core Post Terms Block
+ *
+ * @package WPGraphQL\ContentBlocks\Blocks
+ */
+
+namespace WPGraphQL\ContentBlocks\Blocks;
+
+use WPGraphQL\ContentBlocks\Registry\Registry;
+use WPGraphQL\Model\Term;
+use WP_Block_Type;
+
+/**
+ * Class CorePostTerms.
+ */
+class CorePostTerms extends Block {
+	/**
+	 * {@inheritDoc}
+	 */
+	public function __construct( WP_Block_Type $block, Registry $block_registry ) {
+		parent::__construct( $block, $block_registry );
+
+		$this->register_list_of_terms_field();
+	}
+
+	/**
+	 * Registers a list of terms field for the block.
+	 *
+	 * @return void
+	 * @throws \Exception
+	 */
+	protected function register_list_of_terms_field() {
+		register_graphql_field(
+			$this->type_name,
+			'terms',
+			[
+				'type'        => [ 'list_of' => 'TermNode' ],
+				'description' => __( 'The terms associated with the post.', 'wp-graphql-content-blocks' ),
+				'resolve'     => static function ( $block ) {
+					$term = $block['attrs']['term'] ?? null;
+					if ( empty( $term ) ) {
+						return null;
+					}
+
+					$id = get_the_ID();
+					if ( ! $id ) {
+						return null;
+					}
+
+					$terms = get_the_terms( $id, $term );
+					if ( empty( $terms ) || is_wp_error( $terms ) ) {
+						return null;
+					}
+
+					return array_map( static fn ( $term ) => new Term( $term ), $terms );
+				},
+			]
+		);
+	}
+}
--- a/includes/Blocks/CorePostTerms.php
+++ b/includes/Blocks/CorePostTerms.php
@@ -31,7 +31,7 @@ class CorePostTerms extends Block {
 			]
 		);
 
-		$this->register_list_of_terms_field();
+		$this->register_list_of_terms_connection();
 	}
 
 	/**
@@ -58,7 +58,7 @@ class CorePostTerms extends Block {
 	 * @return void
 	 * @throws \Exception
 	 */
-	protected function register_list_of_terms_field() {
+	protected function register_list_of_terms_connection() {
 		register_graphql_connection(
 			[
 				'fromType'      => $this->type_name,
--- a/includes/Blocks/CorePostTerms.php
+++ b/includes/Blocks/CorePostTerms.php
@@ -9,6 +9,7 @@ namespace WPGraphQL\ContentBlocks\Blocks;
 
 use WPGraphQL\AppContext;
 use WPGraphQL\ContentBlocks\Registry\Registry;
+use WPGraphQL\Data\Connection\TaxonomyConnectionResolver;
 use WPGraphQL\Data\Connection\TermObjectConnectionResolver;
 use WP_Block_Type;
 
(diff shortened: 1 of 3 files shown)


includes/Blocks/CorePostTerms.php
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	public const STRING_FIELDS = [ 'prefix', 'suffix', 'term' ];
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function __construct( WP_Block_Type $block, Registry $block_registry ) {
+		parent::__construct( $block, $block_registry );
+
+		foreach ( self::STRING_FIELDS as $field ) {
```
justlevine, line 29: This seems like an unnecessary abstraction. I'd recommend using `register_graphql_fields()` (plural) directly

includes/Blocks/CorePostTerms.php
```diff
@@ -0,0 +1,97 @@
...
+	 * @throws \Exception
+	 */
+	protected function register_list_of_terms_field() {
+		register_graphql_field(
+			$this->type_name,
+			'terms',
+			[
+				'type'        => [ 'list_of' => 'TermNode' ],
+				'description' => __( 'The terms associated with the post.', 'wp-graphql-content-blocks' ),
+				'resolve'     => static function ( $block ) {
```
justlevine, line 76: `terms` should be a connection, but if there was justification for it not to be, you'd want to use `$context->get_loader( 'term' )->load_deferred( $term_id )` (on my phone, confirm syntax) instead of manually converting it to a Model.

whoami-pwd: @justlevine Thank you for pointing that out!

You are absolutely right:
`array_map( static fn($term) => $context->get_loader('term')->load_deferred($term->term_id), $terms );`
aligns better with WPGraphQL best practices by leveraging deferred loading for efficiency.

Also, great call on using a connection here! I’ve updated the code accordingly 👍🏻

includes/Blocks/CorePostTerms.php
whoami-pwd: @justlevine Good idea. 
I’ve updated the code to remove the extra method call and directly register fields inside `register_graphql_fields()`. The abstraction has been minimized, but `get_string_field_config()` is still used for maintainability

theodesp (approved):
Awesome!

includes/Blocks/CorePostTerms.php
```diff
@@ -0,0 +1,86 @@
...
+
+					$resolver = new TermObjectConnectionResolver( $block, $args, $context, $info, $term );
+					$resolver->set_query_arg( 'object_ids', $post_id );
+
+					return $resolver->get_connection();
+				},
+			]
+		);
+	}
+}
```
justlevine, line 86: 1. There's 0 reason to abstract the fields into a factory. Just makes it harder to read, refactor, and fuzzies out the descrptions.
2. Just because the attribute is stupidly called `term` it's actually a `taxonomyName`, so call it such.
3. ...which means there should be a connection to the taxonomy too.
4. Always better to pass the Connection Resolver values via `$args` than it is to `->set...()` them after the fact.

```suggestion
<?php
<?php
/**
 * Core Post Terms Block
 *
 * @package WPGraphQL\ContentBlocks\Blocks
 */

namespace WPGraphQL\ContentBlocks\Blocks;

use WPGraphQL\AppContext;
use WPGraphQL\ContentBlocks\Registry\Registry;
use WPGraphQL\Data\Connection\TaxonomyConnectionResolver;
use WPGraphQL\Data\Connection\TermObjectConnectionResolver;
use WP_Block_Type;

/**
 * Class CorePostTerms.
 */
class CorePostTerms extends Block {
	/**
	 * {@inheritDoc}
	 */
	public function __construct( WP_Block_Type $block, Registry $block_registry ) {
		parent::__construct( $block, $block_registry );

		$this->register_fields();
		$this->register_connections();
	}

	/**
	 * Registers custom fields for the block.
	 */
	private function register_fields(): void {
		register_graphql_fields(
			$this->type_name,
			[
				'prefix'       => [
					'type'        => 'String',
					'description' => __( 'Prefix to display before the post terms', 'wp-graphql-content-blocks' ),
					'resolve'     => static fn ( $block ) => isset( $block['attrs']['prefix'] ) ? (string) $block['attrs']['prefix'] : null,
				],
				'suffix'       => [
					'type'        => 'String',
					'description' => __( 'Suffix to display after the post terms', 'wp-graphql-content-blocks' ),
					'resolve'     => static fn ( $block ) => isset( $block['attrs']['suffix'] ) ? (string) $block['attrs']['suffix'] : null,
				],
				'taxonomyName' => [
					'type'        => 'String',
					'description' => __( 'The taxonomy name to display terms from', 'wp-graphql-content-blocks' ),
					'resolve'     => static fn ( $block ) => isset( $block['attrs']['term'] ) ? (string) $block['attrs']['term'] : null,
				],
			]
		);
	}

	/**
	 * Registers a list of terms field for the block.
	 *
	 * @return void
	 * @throws \Exception
	 */
	protected function register_connections() {
		// Register connection to terms.
		register_graphql_connection(
			[
				'fromType'      => $this->type_name,
				'toType'        => 'TermNode',
				'fromFieldName' => 'terms',
				'resolve'       => static function ( $block, array $args, AppContext $context, $info ) {
					$taxonomy = $block['attrs']['term'] ?? null;
					if ( empty( $taxonomy ) ) {
						return null;
					}

					$post_id = get_the_ID();
					if ( ! $post_id ) {
						return null;
					}

					$args['where']['objectIds'] = $post_id;

					$resolver = new TermObjectConnectionResolver( $block, $args, $context, $info, $taxonomy );

					return $resolver->get_connection();
				},
			]
		);

		// Register connection to the taxonomy.
		register_graphql_connection(
			[
				'fromType'      => $this->type_name,
				'toType'        => 'Taxonomy',
				'fromFieldName' => 'taxonomy',
				'oneToOne'      => true,
				'resolve'       => static function ( $block, array $args, AppContext $context, $info ) {
					$taxonomy = $block['attrs']['term'] ?? null;

					if ( empty( $taxonomy ) ) {
						return null;
					}

					$resolver = new TaxonomyConnectionResolver( $block, $args, $context, $info );
					$resolver->set_query_arg( 'name', $taxonomy );

					return $resolver->one_to_one()->get_connection();
				},
			]
		);
	}
}
```

whoami-pwd: @justlevine Thank you for your insights - I appreciate the detailed feedback!

- After reviewing it again, especially considering the variations in descriptions, I agree that abstracting just three items here may not provide enough benefit to justify the added complexity of the abstractions itself.

- Good point regarding the naming - calling it `term` could indeed be misleading. I suggest we proceed with `taxonomySlug` instead of `taxonomyName` for better clarity.

- Your note about the taxonomy connection is well taken. I can see how establishing a proper connection will be beneficial, and I’ll make sure to integrate that.

- Thanks also for the suggestion on passing Connection Resolver values via `$args` rather than setting them afterward! I’ll definitely incorporate this approach where applicable.

Appreciate the guidance, and looking forward to refining this further! Let me know if you have any additional thoughts.

whoami-pwd: - ~~Good point regarding the naming - calling it term could indeed be misleading. I suggest we proceed with taxonomySlug instead of taxonomyName for better clarity.~~
I removed the taxonomy string field in favor of using the connection itself.

Issue #136, opened 2023-07-24
`CoreImageAttributes.width` is type `String`, but should be a `Float`


justlevine:
Noticing that the `width` field on `CoreImageAttributes` is type `String` when it should be of [type `Float`](https://github.com/WordPress/gutenberg/blob/70d3e305a662edce554a3408b47da41fabab497a/packages/block-library/src/image/block.json#L66):

![image](https://github.com/wpengine/wp-graphql-content-blocks/assets/29322304/293000f0-1985-4159-973a-e8bc51241e2d)

I'm not 💯sure how this plugin is automating the `{BlockType}Attributes` field registration, otherwise I'd open a PR myself.

theodesp:
Actually this causes an issue if you turn this into a Float when you combine this with CoreColumns

<img width="734" alt="252650454-4bfd991b-edd6-4650-b0c7-59221e37ece5" src="https://github.com/wpengine/wp-graphql-content-blocks/assets/328805/6d29d49d-ef61-4e11-a200-407c60298b3b">

Take a look at this testcase:

https://github.com/wpengine/wp-graphql-content-blocks/blob/main/tests/unit/blocks/CoreImageTest.php#L38

I think my fix was premature. Any ideas or thoughts on how to support this better is welcome.

justlevine:
> I think my fix was premature. Any ideas or thoughts on how to support this better is welcome.

From a [search](https://github.com/search?q=repo%3AWordPress%2Fgutenberg+%22width%22%3A+%7B+%22type%22%3A+%22string%22+path%3A**%2Fblock.json&type=code) of WordPress's codebase, it seems, the Column and Spacer blocks are the only one that use a `string` width, while the rest use a `number`. As such it should be treated as outlier behavior.

As mentioned, I'm unclear where this is being handled in the code and I don't see a particular PR that made this fix, but my suggestion would be to parse the width into a `width: Float, widthUnit: CSSUnitEnum`. This is the pattern used by the [Search block](https://github.com/WordPress/gutenberg/blob/1afe0eae37a5f8bc51d0c87efc82ce7a5c9122b9/packages/block-library/src/search/block.json#L27), and will likely be adopted by the Column and Spacer blocks in a future release, so not just would this be a semantically ideal pattern, it's forward compatible.

theodesp:
[...]

PR #340, opened 2025-02-06
Fix revert coreimage block width type

theodesp:
Reverts https://github.com/wpengine/wp-graphql-content-blocks/pull/130

### DIFF
--- a/.changeset/many-garlics-divide.md
+++ b/.changeset/many-garlics-divide.md
@@ -1,5 +1,7 @@
 ---
-"@wpengine/wp-graphql-content-blocks": patch
+"@wpengine/wp-graphql-content-blocks": major
 ---
 
 Fix: reverts CoreImage block width type introduced at this [PR](https://github.com/wpengine/wp-graphql-content-blocks/pull/130)
+
+**NOTE**: This change is only impactful for users running WordPress versions earlier than v6.3.2.
--- a/.changeset/many-garlics-divide.md
+++ b/.changeset/many-garlics-divide.md
@@ -1,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+Fix: reverts CoreImage block width type introduced at this [PR](https://github.com/wpengine/wp-graphql-content-blocks/pull)
--- a/.changeset/many-garlics-divide.md
+++ b/.changeset/many-garlics-divide.md
@@ -2,4 +2,4 @@
 "@wpengine/wp-graphql-content-blocks": patch
 ---
 
-Fix: reverts CoreImage block width type introduced at this [PR](https://github.com/wpengine/wp-graphql-content-blocks/pull)
+Fix: reverts CoreImage block width type introduced at this [PR](https://github.com/wpengine/wp-graphql-content-blocks/pull/130)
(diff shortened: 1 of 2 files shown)


.changeset/many-garlics-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wpengine/wp-graphql-content-blocks": patch
+---
+
+Fix: reverts CoreImage block width type introduced at this [PR](https://github.com/wpengine/wp-graphql-content-blocks/pull)
```
colinmurphy, line 5:
```suggestion
Fix: reverts CoreImage block width type introduced at this [PR](https://github.com/wpengine/wp-graphql-content-blocks/pull/130)
```

colinmurphy (dismissed):

colinmurphy (dismissed):

colinmurphy:
@theodesp LTGM

justlevine:
> Latest Image block.json width is of type "string" so I don' think this is a breaking change
> 
> https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks/image/block.json#L74 https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks/image/block.json#L74

Yes it is...

image

theodesp:
> > Latest Image block.json width is of type "string" so I don' think this is a breaking change
> > https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks/image/block.json#L74 https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks/image/block.json#L74
> 
> Yes it is...
> 
> image

Yes. I've seen this. The thing is, I don't see where this `width: "number"` type is coming from? All the references of the image width in wordpress latest or develop are "width: "string"

Also after trying to query the Schema with this PR with WP v6.7.1, I can see the type is string:

Screenshot 2025-02-10 at 12 00 04

Is it the `schema.graphql` out of date maybe?

justlevine:
> > > Latest Image block.json width is of type "string" so I don' think this is a breaking change
> > > https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks/image/block.json#L74 https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks/image/block.json#L74
> > 
> > 
> > Yes it is...
> > image
> 
> Yes. I've seen this. The thing is, I don't see where this `width: "number"` type is coming from? All the references of the image width in wordpress latest or develop are "width: "string"
> 
> Also after trying to query the Schema with this PR with WP v6.7.1, I can see the type is string:
> 
> Screenshot 2025-02-10 at 12 00 04
> Is it the `schema.graphql` out of date maybe?

My understanding of your CI, is that you're defining the WordPress version in the [.env.dist](https://github.com/wpengine/wp-graphql-content-blocks/blob/8bd5bd40b2c9d5375ae01ddeb68b5dac9b57118a/.env.dist#L52) instead of in the CI itself, so makes sense it would get out dated, but that's not the issue here.

The issue is that the minimum WP requirement isn't 6.7, and as such removing this is a breaking change for users on earlier versions of WordPress.

(Personally, I'd much rather break this just once as part of #136 / interfaces that stabilize the schema independently underlying changes to WordPress )

PR #377, opened 2025-04-22
Add WPGraphQL version compatibility headers and checks

theodesp:
# Description
This PR introduces custom plugin headers for declaring the minimum required and maximum tested versions of WPGraphQL, and wires them into the plugin's initialization logic. This provides better validation and safer bootstrapping when used alongside WPGraphQL.

Added the following custom headers to the main plugin file:

`Requires WPGraphQL`

`WPGraphQL tested up to`

## How to Test
✅ Ensure WPGraphQL is installed and activated.

✅ Change the version of WPGraphQL (if needed) to simulate:

* A version lower than Requires WPGraphQL

* A version higher than WPGraphQL tested up to

✅ Activate the plugin:

If the WPGraphQL version is below the required, the plugin should not continue loading and a WP admin notice should appear.

If within the compatible range, the plugin should load normally.

✅ Check admin notice behavior:

Go to `WP Admin > Plugins` and verify the message when versions are incompatible.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,6 @@
   "require": {
     "php": ">=7.4",
     "imangazaliev/didom": "^2.0",
-    "blakewilson/wp-enforce-semver": "^3.0",
     "cweagans/composer-patches": "^1"
   },
   "require-dev": {
@@ -20,7 +19,8 @@
     "szepeviktor/phpstan-wordpress": "^1.3",
     "axepress/wp-graphql-stubs": "^1.14",
     "axepress/wp-graphql-cs": "^2.0.0-beta",
-    "roave/security-advisories": "dev-latest"
+    "roave/security-advisories": "dev-latest",
+    "wp-graphql/wp-graphql": "^2.2"
   },
   "scripts": {
     "install-test-env": "bash bin/install-test-env.sh",
--- a/includes/WPGraphQLContentBlocks.php
+++ b/includes/WPGraphQLContentBlocks.php
@@ -102,13 +102,6 @@ final class WPGraphQLContentBlocks {
 			require_once WPGRAPHQL_CONTENT_BLOCKS_PLUGIN_DIR . '/includes/Updates/CheckForUpgrades.php';
 		}
 
-		// Bail if the Enforce SemVer class doesn't exist.
-		if ( ! class_exists( 'EnforceSemVer\EnforceSemVer' ) ) {
-			return false;
-		}
-
-		new \EnforceSemVer\EnforceSemVer( WPGRAPHQL_CONTENT_BLOCKS_PLUGIN_PATH );
-
 		return $success;
 	}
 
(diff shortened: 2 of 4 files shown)


colinmurphy:
@theodesp 

We should also remove and the corresponding patch.

```
  "extra": {
    "patches": {
      "blakewilson/wp-enforce-semver": {
        "Fixes version key case": "patches/blakewilson-wp-enforce-semver/semver-object-key-version.patch"
      }
    }
  }
```

composer.json
```diff
@@ -20,7 +19,8 @@
     "szepeviktor/phpstan-wordpress": "^1.3",
     "axepress/wp-graphql-stubs": "^1.14",
     "axepress/wp-graphql-cs": "^2.0.0-beta",
-    "roave/security-advisories": "dev-latest"
+    "roave/security-advisories": "dev-latest",
+    "wp-graphql/wp-graphql": "^2.2"
```
justlevine, line 23: Do we need WPGraphQL as its own dev-dep? If you're trying to lint/typecheck against a specific version then wp-graphql-stubs without all the potential dependency conflicts.

includes/WPGraphQLContentBlocks.php
```diff
@@ -119,6 +110,38 @@ static function () {
...
+		$this->register_wpgraphql_update_filters();
+	}
+
+	/**
+	 * Register filters related to WPGraphQL version compatibility.
+	 *
+	 * @since 0.0.1
+	 */
+	public function register_wpgraphql_update_filters(): void {
+		add_filter( 'graphql_get_dependents', [ $this, 'add_as_wpgraphql_dependent' ], 10, 2 );
```
justlevine, line 122: Since you're using the header, you dont need to force-add yourself using a filter.

includes/WPGraphQLContentBlocks.php
```diff
@@ -119,6 +110,38 @@ static function () {
...
+
+		add_filter( 'wpgraphql_enable_untested_autoupdates', function ($allow, $release_type) {
```
justlevine, line 124: This filter is for whether _WPGraphQL_ should autoupdate for any plugin, not just this one.

But also, untested autoupdates are allowed by default for non _major_ versions: https://github.com/wp-graphql/wp-graphql/blob/98457dc6a3412e34ba2f6cbeef848982904ab7a8/src/Admin/Updates/UpdateChecker.php#L352

composer.json
theodesp: I will remove

includes/WPGraphQLContentBlocks.php
theodesp: Does it work without the filter though? Just keeping the following lines in `wp-graphql-content-blocks.php`

```
 * Requires WPGraphQL:       1.14.5
 * WPGraphQL tested up to:   2.1.1
 ```
 should be enough?

justlevine: I mean I'm not sure if WordPress (what wpgraphql) uses underneath is able to handle the whitespace formatting, but yeah that's enough.

colinmurphy (approved):

PR #442, opened 2026-05-26
chore(deps): patch uuid and webpack-dev-server security alerts (dev-only)

josephfusco:
## Summary

Adds two entries to the `overrides` block in `package.json` to force transitive dependencies to patched versions:

- **`uuid` → `^11.1.1`** — resolves GHSA-w5hq-g745-h8pq (missing buffer bounds check in v3/v5/v6)
- **`webpack-dev-server` → `^5.2.4`** — resolves GHSA-79cf-xcqc-c78w (cross-origin source code exposure on non-HTTPS)

Dependabot couldn't auto-fix either: `uuid` is purely transitive (Dependabot's job is configured `allowed-updates: direct` only — 4 prior runs failed), and `webpack-dev-server` was pinned at 4.x via `@wordpress/scripts`'s constraint with no 4.x backport from upstream.

## Why this is non-breaking for consumers

Neither package is referenced by code in this repo (no `require`/`import` of `uuid` or `webpack-dev-server` anywhere in `includes/`, `bin/`, or root `*.php`). Both exist only in the dev/test tooling tree. The plugin's `files` array publishes only `includes/`, `vendor/`, license/readme, and `wp-graphql-content-blocks.php` — `node_modules` never ships.

## Test plan

- [x] `npm install` succeeds with no peer-dep conflicts (409 added, 181 removed, 537 changed)
- [x] `npm ls uuid` returns `uuid@11.1.1 overridden` (single resolution)
- [x] `npm ls webpack-dev-server` returns `webpack-dev-server@5.2.4 overridden` (single resolution)
- [x] `@wordpress/scripts` and `webpack-dev-server` both `require()` cleanly under Node 20
- [ ] CI: existing PHPStan, WPCS, Schema Linter, and Test Plugin workflows pass
- [ ] After merge: dependabot alert status for both GHSAs flips to resolved/closed

## Release impact

Includes a `patch`-level changeset so this rolls into the next release (currently `@wpengine/wp-graphql-content-blocks@4.8.5` queued in #403).

PR #443, opened 2026-05-27
fix(deps): add missing webpack-dev-server override (unbreaks main)

josephfusco:
## Summary

- `npm ci` on `main` is failing — PR #442 (1c3b6df) added the `webpack-dev-server@5.2.4` resolution to `package-lock.json` but left the corresponding entry **out** of the `overrides` block in `package.json`. The lockfile has `5.2.4` while package.json's natural tree (via `@wordpress/scripts`) wants `4.15.2`, with no override to bridge them.
- This breaks the `release.yml` workflow (failed run [26516976225](https://github.com/wpengine/wp-graphql-content-blocks/actions/runs/26516976225)) — without it, the changesets action can't install deps to refresh PR #403 or publish 4.8.5.
- Adding the missing `"webpack-dev-server": "^5.2.4"` line is sufficient; lockfile is already at the correct resolution and `npm install --package-lock-only` is a no-op.

## Test plan

- [x] `npm ci --dry-run` exits 0 locally with the patched `package.json`
- [x] `npm install --package-lock-only` makes no further changes to the lockfile
- [ ] CI green on this PR
- [ ] After merge: release.yml succeeds, PR #403 refreshes with the security changeset alongside WP 6.9 compat

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   "overrides": {
     "serialize-javascript": "^7.0.5",
     "uuid": "^11.1.1",
+    "webpack-dev-server": "^5.2.4",
     "@typescript-eslint/typescript-estree": {
       "minimatch": "^9.0.7"
     },


josephfusco:
Review nudge for @wpengine/headless-open-source — this is a **hotfix to unbreak `main`** after #442.

PR #442's lockfile shipped `webpack-dev-server@5.2.4` but the corresponding override line was dropped from `package.json` during merge. `npm ci` on main is now failing with `Missing: webpack-dev-server@4.15.2 from lock file`, which broke the release.yml run triggered by the #442 merge (run [26516976225](https://github.com/wpengine/wp-graphql-content-blocks/actions/runs/26516976225)).

This adds the one missing line. Verified locally: `npm ci --dry-run` exits 0; `npm install --package-lock-only` is a no-op (lockfile is already correct).

Once this is in, release.yml will run again, refresh #403 with the security changeset, and we can ship 4.8.5.

ahuseyn (approved):

